PR #18, opened 2021-06-09
CI: Pretty-print JUnit report

nomeata:


nomeata:
Playing around with https://github.com/marketplace/actions/publish-unit-test-results. But it looks complicated…

nomeata:
Not sure how useful it is in the end, but it certainly looks pretty…

nomeata:
Oh, this isn't from my own repo. Will have to reopen.

PR #22, opened 2021-07-12
ic-ref-run: Pretty print candid values without type table

### LINKED ISSUE #21: Fix pretty-printing of messages in `ic-ref-run`
https://github.com/dfinity/motoko/pull/2654 shows that `ic-ref-run` pretty-prints the message contents wrongly; this is a consequence of https://github.com/nomeata/haskell-candid/commit/01157c50ec29b1f8ab35ca27f38dc592ed5406f9#diff-a70b08fdc8d7e85a721cc42160181f1c12dcfedbe6c08d298fd5f5ac71f24097R27 changing the type signature of `decodeVals` to also return the type table. A `snd` in the right spot should fix this.

nomeata:
fixes #21


ggreif (approved):
Looks reasonable, let's try this.

Issue #24, opened 2021-07-14
Cannot build Serialise.hs with GHC 8.10.4


jwiegley:
Using the latest nixpkgs-unstable and GHC 8.10.4, I get the following error during compilation:
```
[49 of 57] Compiling IC.Serialise     ( src/IC/Serialise.hs, dist/build/IC/Serialise.o, dist/build/IC/Serialise.dyn_o )

src/IC/Serialise.hs:25:1: error:
    Could not find module ‘System.Random.Internal’
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
   |
25 | import System.Random.Internal (StdGen(..))
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Pinging @nomeata

nomeata:
Hmm, that rings a bell, maybe I addressed that in one of the PRs already? Let me check...

Maybe the whole Serialize.hs module should be removed again. This was created for the use of ic-ref as an emulator in dfx (dfx start --emulator), where @hansl said that it’s important that the state can be persisted. I was never convinced, and I think Hans wants to write a separate emulator anyways. Hans?
 
I can’t find the branch where I tried to address this, if it ever existed. One solution is certainly to continue using an older version of random, where StdGen is not abstract. Another is to use the splitmix library directly, and store `(Word64, Word64)` in the state, and use `seedSMGen'` and `unseedSMGen` to produce the SMGen.

Something else must have stopped me while I tried to fix this back then. Maybe the problem with building rust on the latest nixpkgs.

nomeata:
Ah, I thought PRs #11 or #19 would actually upgrade GHC as well, but it seems that 8.4 is still used there. Let’s see what I can do.

nomeata:
Ok, something is odd. We have
```
  build-depends: random >= 1.2
```
and the only version of random with that constraint is [1.2.0](https://hackage.haskell.org/package/random-1.2.0/random.cabal) which exports `System.Random.Internal`.

nomeata:
Hmpf. That was a relative long time to find out that I can’t reproduce this. On current `master`, with `cabal.project.freeze` removed and all version bounds on `base` removed from `ic-hs.cabal`, I can run
```
~/dfinity/ic-hs $ cabal build -w ghc-8.10.4 
```
works just fine for me.

@jwiegley, what revision are you on, what changed did you make, and what is your precise command?

nomeata:
[...]

PR #19, opened 2021-06-16
nix: upgrade nixpkgs to the latest release (21.05)

basvandijk:
This is an experiment to see if this solves the linking errors on MacOS Big Sur.

PR #25, opened 2021-07-15
Update freeze file for GHC-8.10.4

nomeata:
I tried to reproduce #24, but couldn’t. This is the freeze file that I got out of using ghc-8.10.4

Not to be merged like this. Maybe once we update nixpkgs and 8.10 is the default GHC (see #19)


nomeata:
Subsumed by #30

PR #26, opened 2021-07-20
Adds verification of RSA signatures

kritzcreek:


src/IC/Crypto/WebAuthn.hs
```diff
@@ -72,94 +77,126 @@ genClientDataJson challenge = JSON.encode $ JSON.Object $
...
-
+genCOSEECDSAKey (EC.PublicKey _ EC.PointO) = error "genCOSEKey: Point at infinity"
 
+genCOSERSAKey :: RSA.PublicKey -> BS.ByteString
+genCOSERSAKey (RSA.PublicKey keyLength n e) =
+  toLazyByteString $ encodeTerm $ TMap
+    [ (TInt 1, TInt 3)
+    , (TInt 3, TInt (-257))
+    , (TInt (-1), TBytes (EC.i2ospOf_ keyLength n))
+    , (TInt (-2), TBytes (EC.i2ospOf_ keyLength e))
```
kritzcreek, line 138:
```suggestion
    , (TInt (-1), TBytes (EC.i2osp n))
    , (TInt (-2), TBytes (EC.i2osp e))
```

From section 4: https://datatracker.ietf.org/doc/html/rfc8230#section-4
> All numeric key parameters are encoded in an unsigned big-endian
      representation as an octet sequence using the CBOR byte string
      type (major type 2).  The octet sequence MUST utilize the minimum
      number of octets needed to represent the value.

src/IC/Crypto/WebAuthn.hs
```diff
@@ -72,94 +77,122 @@ genClientDataJson challenge = JSON.encode $ JSON.Object $
...
+        e <- integerField (-2)
+        let pk = RSA.PublicKey 256 n e
+        unless (RSA.verify (Just SHA256) pk (BS.toStrict msg) (BS.toStrict sig)) $
+          throwError "WebAuthn signature verification failed"
+      _ -> throwError $ "COSE: Unsupport pair of unsupported type, algorithm: " <> tshow ty <> " " <> tshow alg
   where
     keyVal (TInt k,v) = pure (fromIntegral k,v)
     keyVal (TInteger k,v) = pure (k,v)
     keyVal _ = throwError "Non-integer key in CBOR map"
+    tshow :: (Show a) => a -> T.Text
```
nomeata, line 118:
```suggestion
    tshow :: Show a => a -> T.Text
```

src/IC/Crypto/WebAuthn.hs
```diff
@@ -72,94 +77,122 @@ genClientDataJson challenge = JSON.encode $ JSON.Object $
...
-
+genCOSEECDSAKey (EC.PublicKey _ EC.PointO) = error "genCOSEKey: Point at infinity"
 
+genCOSERSAKey :: RSA.PublicKey -> BS.ByteString
+genCOSERSAKey (RSA.PublicKey _keyLength n e) =
+  toLazyByteString $ encodeTerm $ TMap
+    [ (TInt 1, TInt 3)
+    , (TInt 3, TInt (-257))
+    , (TInt (-1), TBytes (EC.i2osp n))
+    , (TInt (-2), TBytes (EC.i2osp e))
```
nomeata, line 138: You probably want to use `EC.i2ospOf_ 32` as above, as otherwise I fear it might do the wrong things if there are leading zeros.

nomeata (commented):
Phone review, might have missed stuff. Most prominently lacking is `ic-ref-test` use of the new scheme.

src/IC/Crypto/WebAuthn.hs
kritzcreek: I commented on these lines earlier:

From section 4: https://datatracker.ietf.org/doc/html/rfc8230#section-4

    All numeric key parameters are encoded in an unsigned big-endian
    representation as an octet sequence using the CBOR byte string
    type (major type 2). The octet sequence MUST utilize the minimum
    number of octets needed to represent the value.


According to https://hackage.haskell.org/package/cryptonite-0.29/docs/Crypto-PubKey-RSA.html#v:generateWith common choices for `e` are `0x10001` or `3` 

which makes me doubt 32bytes "utilizes the minimum number of octets needed to represent the value"

src/IC/Crypto.hs
```diff
@@ -41,7 +41,7 @@ createSecretKeyEd25519 :: BS.ByteString -> SecretKey
 createSecretKeyEd25519 = Ed25519 . Ed25519.createKey
 
 createSecretKeyWebAuthn :: BS.ByteString -> SecretKey
-createSecretKeyWebAuthn = WebAuthn . WebAuthn.createKey
+createSecretKeyWebAuthn = WebAuthn . WebAuthn.createECDSAKey
```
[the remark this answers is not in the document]
kritzcreek: My [Fairbairn Threshold](https://wiki.haskell.org/Fairbairn_threshold) detector went off, so I didn't.

But you're right, I'll add the definition for uniformity.

src/IC/Test/WebAuthn.hs
```diff
@@ -26,17 +26,32 @@ webAuthnTests :: TestTree
 webAuthnTests = testGroup "WebAuthn crypto tests"
   [ testProperty "create-sign-verify" $
```
ggreif, line 27:
```suggestion
  [ testProperty "ECDSA: create-sign-verify" $
```
ditto below

src/IC/Crypto/WebAuthn.hs
```diff
@@ -72,94 +77,122 @@ genClientDataJson challenge = JSON.encode $ JSON.Object $
...
+        sig <- parseCOSEECDSASig sig
+        unless (EC.verify SHA256 pk sig (BS.toStrict msg)) $
+            throwError "WebAuthn signature verification failed"
+      (3, -257) -> do
+        n <-  integerField (-1)
+        e <- integerField (-2)
+        let pk = RSA.PublicKey 256 n e
+        unless (RSA.verify (Just SHA256) pk (BS.toStrict msg) (BS.toStrict sig)) $
+          throwError "WebAuthn signature verification failed"
+      _ -> throwError $ "COSE: Unsupport pair of unsupported type, algorithm: " <> tshow ty <> " " <> tshow alg
```
ggreif, line 113: "Unsupport" sounds like a verb here, is that your intention?

kritzcreek:
@nomeata I think I adressed your feedback. PTAL

stefan-at-dfinity (approved):
lgtm

src/IC/Crypto/WebAuthn.hs
```diff
@@ -72,94 +77,122 @@ genClientDataJson challenge = JSON.encode $ JSON.Object $
     <> "type" JSON..= ("webauthn.get" :: T.Text)
     <> "origin" JSON..= ("ic-ref-test" :: T.Text)
 
-parseCOSEKey :: BS.ByteString -> Either T.Text EC.PublicKey
-parseCOSEKey s = do
+verifyCOSESig :: BS.ByteString -> BS.ByteString -> BS.ByteString -> Either T.Text ()
```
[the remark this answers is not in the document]
kritzcreek: (Sorry for the late response, I was on PTO the last two days)

This matches the already exported signature of `verify`, and it's a module internal function. If we wanted to go with better names we should probably start at `verify`.

I think part of the reason we're dealing with lots of ByteStrings in ic-ref's API is that it's primarily a tool used for testing, so it wants to be low-friction for putting together a couple of testcases, rather than having to construct via various newtype wrappers.

PR #29, opened 2021-09-14
Bump haskell-candid

nomeata:
better pretty-printing, among other things


nomeata:
This fails because nixpkgs doesn’t have `prettyprinter-1.7` yet. Maybe wait for that (via https://github.com/dfinity/ic-hs/pull/19?)

nomeata:
Ok, this should be good to go.

nomeata:
I can’t assign reviewers, so maybe manually pinging @marcin-dziadus helps :-)

marcin-dziadus (approved):

PR #31, opened 2021-09-14
Implement heartbeat functionality

marcin-dziadus:


src/IC/Canister.hs
```diff
@@ -29,6 +29,7 @@ type WasmState = CanisterSnapshot
 type InitFunc = EntityId -> Env -> Blob -> TrapOr (WasmState, CanisterActions)
 type UpdateFunc = WasmState -> TrapOr (WasmState, UpdateResult)
 type QueryFunc = WasmState -> TrapOr Response
+type HeartbeatFunc = Env -> WasmState -> TrapOr (UpdateResult)
```
nomeata, line 32: You probably want
```suggestion
type HeartbeatFunc = Env -> WasmState -> TrapOr (WasmState, UpdateResult)
```
because heartbeats need to be state changing

nomeata (commented):
I started having a look, but I guess this is WIP, so stopped. But let me know if you have questions!

src/IC/Canister.hs
marcin-dziadus: Good point, thanks!

marcin-dziadus:
> I started having a look, but I guess this is WIP, so stopped. But let me know if you have questions!

Could you please take a look anyway and check if a general direction seems promising?

src/IC/Test/Universal.hs
```diff
@@ -219,6 +222,9 @@ instance IsString (Exp 'B) where
 callback :: Prog -> Exp 'B
 callback = bytes . run
 
+heartbeat :: Prog -> Exp 'B
+heartbeat = bytes . run
```
nomeata, line 226: Maybe use `callback`, with or without new name here

src/IC/Test/Spec.hs
```diff
@@ -1034,6 +1034,22 @@ icTests = withAgentConfig $ testGroup "Interface Spec acceptance tests"
...
+  , testGroup "heartbeat"
+    [ testCase "called once for all canisters" $ do
+      cid <- install $ onHeartbeat $ heartbeat $ ignore (stableGrow (int 1)) >>> stableWrite (int 0) "FOO"
+      cid2 <- install $ onHeartbeat $ heartbeat $ ignore (stableGrow (int 1)) >>> stableWrite (int 0) "BAR"
+      -- Heartbeat cannot respond. Should be rejected.
+      cid3 <- install $ onHeartbeat $ heartbeat $ setGlobal "FIZZ" >>> replyData "FIZZ"
+
+      -- Send a dummy request to enforce execution of heartbeat functions.
+      -- (heartbeats are triggered for all running canisters after each call request)
+      call_ cid (setCertifiedData "DUMMY" >>> reply)
```
nomeata, line 1046:
```suggestion
      -- The spec currently gives no guarantees about when or how frequent heartbeats are executed.
      -- But all implementation have this property: If update call B is submitted after call A is completed,
      -- then a heartbeat runs before the execution of B.
      -- We use this here to make sure that callbacks have been attempted:
      call_ cid reply
      call_ cid reply
```

marcin-dziadus: Done, thanks!

src/IC/Canister/Imp.hs
```diff
@@ -550,6 +551,25 @@ rawInitialize caller env dat (ImpState esref inst sm wasm_mod) = do
...
+    then withES esref es $ void $ invokeExport inst "canister_heartbeat" []
+    else return ((), es)
+
+  case result of
+    Left  err -> return $ Trap err
+    Right (_, es')
+      | accepted es'          -> return $ Trap "cannot accept_message here"
+      | isJust (response es') -> return $ Trap "cannot respond from heartbeat"
+      | otherwise             -> return $ Return $
+        ( CallActions (calls es') (cycles_accepted es') (response es')
```
nomeata, line 569: `cycles_accepted` should always be 0, becuase `accept_cycles` should have trapped in the heartbeat (does it?). So maybe
```suggestion
        ( CallActions (calls es') 0 Nothing
```

An alternative would be to add the calls to the `CanisterActions`, and have
```
 type HeartbeatFunc = Env -> WasmState -> TrapOr (WasmState, CanisterActions)
 ```
 which is more honest than using `UpdateResult` here. Or 
 ```
 type HeartbeatFunc = Env -> WasmState -> TrapOr (WasmState, CanisterActions, Calls)
 ```

File: src/IC/HTTP.hs [hunk 5]
```diff
@@ -82,6 +82,8 @@ handle store req respond = case (requestMethod req, pathInfo req) of
       x <- modifyStore store $ do
         -- Here we make IC.Ref use “real time”
         lift getTimestamp >>= setAllTimesTo
+        -- Enqueue heartbeat messages for all canisters
+        enqueueHeartbeats
```
nomeata, line 86: I wonder if it makes running `ic-ref-test` much slower. That installs a lot of canisters (at least one per test). Previously that was fine, because unused canisters don’t do much. But now it means a lot of churn, going through each of the hundrets canisters, with each request… So not sure if we want it to be like this for long.

src/IC/Canister/Imp.hs
marcin-dziadus: > `cycles_accepted` should always be 0, becuase `accept_cycles` should have trapped in the heartbeat (does it?). 

That's correct. I changed the type accordingly.

src/IC/HTTP.hs
marcin-dziadus: Indeed, that's a good point. One easy solution I can think of is to ensure that heartbeats are not scheduled more often than every x seconds.

nomeata: That would break your assumption that they are executed between two consecutive calls to the canister. Tricky.

src/IC/Canister/Imp.hs
```diff
@@ -566,7 +566,7 @@ rawHeartbeat env (ImpState esref inst sm wasm_mod) = do
       | accepted es'          -> return $ Trap "cannot accept_message here"
       | isJust (response es') -> return $ Trap "cannot respond from heartbeat"
       | otherwise             -> return $ Return $
-        ( CallActions (calls es') (cycles_accepted es') (response es')
+        ( calls es'
```
nomeata, line 569: Much better :-)

src/IC/HTTP.hs
marcin-dziadus: Ah, you're totally right. Are you okay with leaving a TODO comment here and fixing it slightly later?

nomeata: Sure! I'm not using ic-ref myself these days, so this itch won't itch me directly (besides maybe CI times).

It would be good to know how long running the test suite takes before and after, to know how bad it actually is.

marcin-dziadus: It gets an order of magnitude slower... I think it's way too much.

src/IC/Ref.hs
```diff
@@ -1167,10 +1170,17 @@ newHeartbeat cid = do
...
+  is_empty   <- isCanisterEmpty cid
+  is_running <- isCanisterRunning cid
+  return $ not is_empty && is_running
+
 enqueueHeartbeats :: ICM m => m ()
 enqueueHeartbeats = do
   cs <- gets (M.keys . canisters)
-  forM_ cs newHeartbeat
+  cs_with_hb <- filterM hasHeartbeat cs
+  forM_ cs_with_hb newHeartbeat
```
nomeata, line 1183: Hmm, is this correct? What if you enqueue a heartbeat call now, but the queue contains a call that will stop or delete the canister?

src/IC/HTTP.hs
nomeata: Ok, then maybe this would work:

* Don't enqueue heartbeat. Simply have a function that goes through all canisters and runs the callback if needed, but keep it separate from the queues (which seems odd anyways)
* with the canister, remember if the last message executed was a heartbeat. Skip canisters where nothing happened since the last heartbeat.

That might run more smoothly.

Alternatively:
* Run a heartbeat before each message execution for just that canister.

marcin-dziadus: I implemented the first variant and now performance overhead is negligible.

src/IC/Ref.hs
marcin-dziadus: It's a very good point, thanks!

marcin-dziadus: The PR is finally green. I'd appreciate one more round of review :)

marcin-dziadus: I think it should be correct now. (at least the PR got green) 
I'd appreciate one more round of review. :)

src/IC/Ref.hs
```diff
@@ -136,14 +137,16 @@ data CanState = CanState
   , time :: Timestamp
   , cycle_balance :: Natural
   , certified_data :: Blob
+  , last_action :: Maybe EntryPoint
```
nomeata, line 140: This might deserve a comment. 
```suggestion
  , last_action :: Maybe EntryPoint
   -- ^ Not part of the spec, but in this implementation we schedule
         heartbeats only for canisters who “did something else” since the
         last heartbeat, so we remember the last action.
         
```

src/IC/Ref.hs
```diff
@@ -705,13 +714,15 @@ processMessage m = case m of
...
         Return (new_state, (call_actions, canister_actions)) -> do
+          modCanister callee $ \cs -> cs { last_action = Just entry }
           performCallActions ctxt_id call_actions
           performCanisterActions callee canister_actions
           setCanisterState callee new_state
 
   ResponseMessage ctxt_id response refunded_cycles -> do
     ctxt <- getCallContext ctxt_id
     case origin ctxt of
+      FromHeartbeat -> return ()
```
nomeata, line 725: This can’t happen, right? 
```suggestion
      FromHeartbeat -> error "Response from heartbeat"
```

src/IC/Ref.hs
```diff
@@ -808,6 +819,7 @@ invokeManagementCanister caller ctxt_id (Public method_name arg) =
         Right x -> method (raw_reply . encode @b) x
 
 invokeManagementCanister _ _ Closure{} = error "closure invoked on management function "
+invokeManagementCanister _ _ Heartbeat = error "heartbeat invoked on management function "
```
nomeata, line 822:
```suggestion
invokeManagementCanister _ _ Closure{} = error "closure invoked on management canister"
invokeManagementCanister _ _ Heartbeat = error "heartbeat invoked on management canister"
```

src/IC/Ref.hs
```diff
@@ -1142,6 +1158,33 @@ setAllTimesTo :: ICM m => Timestamp -> m ()
 setAllTimesTo ts = modify $
   \ic -> ic { canisters = M.map (\cs -> cs { time = ts }) (canisters ic) }
 
+runHeartbeat :: ICM m => CanisterId -> m ()
+runHeartbeat cid = do
+  can <- getCanister cid
+  if idleSinceLastHeartbeat (last_action can)
+  then return ()
+  else do
```
nomeata, line 1166:
```suggestion
  unless (idleSinceLastHeartbeat (last_action can)) $ do
```

src/IC/Ref.hs
```diff
@@ -1142,6 +1158,33 @@ setAllTimesTo :: ICM m => Timestamp -> m ()
...
+    new_ctxt_id <- newCallContext $ CallContext
+      { canister = cid
+      , origin = FromHeartbeat
+      , responded = Responded False
```
nomeata, line 1170: I wonder if these call contexts should be considered “responded”. After all, we don’t expect a response from them.
```suggestion
      , responded = Responded True
```

src/IC/Test/Spec.hs
```diff
@@ -603,6 +603,12 @@ icTests = withAgentConfig $ testGroup "Interface Spec acceptance tests"
             cid <- install (onInspectMessage (callback (prog >>> acceptMessage)))
             call'' cid reply
           )
+        , "H" =: boolTest (\prog -> do
+            cid <- install (onHeartbeat (callback (prog >>> setGlobal "Did not trap")))
+            call_ cid reply
```
nomeata, line 608:
```suggestion
            call_ cid reply -- This assumes that after one update call returned, a heartbeat
                            -- should have happened. Also see heartbeat tests below.
```

src/IC/Test/Spec.hs
```diff
@@ -2417,7 +2443,10 @@ isRelay = runGet $ Get.getWord32le >>= \case
 -- Shortcut for test cases that just need one canister
```
nomeata, line 2443:
```suggestion
-- Shortcut for test cases that just need one canister.
-- Deletes the canister afterwards (so don’t use it for tests that
-- delete canisters, or change the controller)
```

nomeata (approved):
Did another pass. Overall good.

I wonder if deleting canisters in `simpleTestCase` is actually a win, given that it’s an extra 2-3s per test to wait. But 🤷🏻 .

When I worked on these things, and the replica already implemented the feature, I usually tested my ic-hs branch and latest replica. That was pretty easy when we were still using nix (just use `NIV_OVERRIDE`), but I have no idea how you would do that kind of testing now. Worth finding out and documenting, though!

I would be annoying to not test that now, then merge this, maybe even make a tagged release together with the spec, bump it in the `dfinity` repo, and _then_ notice that the `ic-ref-test` tests don’t work well together with the replica.

nomeata:
CI failure due to https://github.com/dfinity/ic-hs/pull/35 likely, so let’s land that first

src/IC/Ref.hs
marcin-dziadus: Right, thanks!

src/IC/Ref.hs
nomeata: This might also be relevant for the transition from stopping to stopped.

src/IC/Test/Spec.hs
marcin-dziadus: Indeed, we changed the semantics in the meantime. Thanks!

marcin-dziadus:
> Did another pass. Overall good.
> 
> I wonder if deleting canisters in `simpleTestCase` is actually a win, given that it’s an extra 2-3s per test to wait. But 🤷🏻 .
> 
> When I worked on these things, and the replica already implemented the feature, I usually tested my ic-hs branch and latest replica. That was pretty easy when we were still using nix (just use `NIV_OVERRIDE`), but I have no idea how you would do that kind of testing now. Worth finding out and documenting, though!
> 
> I would be annoying to not test that now, then merge this, maybe even make a tagged release together with the spec, bump it in the `dfinity` repo, and _then_ notice that the `ic-ref-test` tests don’t work well together with the replica.

That's a perfect point! I would be annoyed, too.

src/IC/Ref.hs
marcin-dziadus: Hmm... not sure. Please look at the comment down below.

src/IC/Ref.hs
marcin-dziadus: @nomeata Then the check in l. 634 is violated.
Heartbeat is invoked through `invokeEntry`, as a result a `ResponseMessage` is enqueued  in l. 640, then simply omitted in l. 725. It seems slightly indirect, but correct. What do you think?

src/IC/Ref.hs
nomeata: Sorry, I misformated the comment. Fancy inline error message from GitHub! Although incomplete.

src/IC/Ref.hs
nomeata: Hmm, but why is there even a response message generated when the canister can't `reply`?

My theory: the call context looks starved (because of `Responded False` and no callbacks), and then the general starvation rule creates the response message. But on the phone right now, which means I may miss the full picture. 

In any case it isnt seemly for a reference implementation to do something such indirect, I'd say, if possible. Maybe there simply shouldn't be a call context when there is, after all, no call involved. Or we need different contexts. Not sure right now.

marcin-dziadus: Thanks, I'll investigate further!

nomeata: There should never be a `ResponseMessage` with a Hearbeat call context. I suggest to add  to `respondCallContext` a check
```
  when (origin ctxt == Heartbeat) $
    error "Internal error: Heartbeats cannot be responded to"
```
to fail early (similar to the double reply check there).

And I don’t see how `invokeEntry` would generate a response there. I think it was due to `starveCallContext` maybe.

src/IC/Ref.hs
```diff
@@ -1056,6 +1071,10 @@ invokeEntry ctxt_id wasm_state can_mod env entry = do
                         Return (wasm_state', (noCallActions, noCanisterActions))
                 Nothing -> Trap err
             Return (wasm_state, actions) -> Return (wasm_state, actions)
+      Heartbeat -> return $ do
+        case heartbeat can_mod env wasm_state of
+            Trap _ -> Return (wasm_state, (noCallActions, noCanisterActions))
+            Return (wasm_state, (calls, actions)) -> Return (wasm_state, (CallActions calls 0 Nothing, actions))
```
nomeata, line 1077: This is  a bit neater, putting emphasis on what’s there (instead of what’s not there), and robust against additions to the record.
```suggestion
            Return (wasm_state, (calls, actions)) -> Return (wasm_state, (noCallActions { ca_new_calls = calls }, actions))
```

src/IC/Ref.hs
marcin-dziadus: Yes, `invokeEntry` has nothing to do with this. The thing is that we're attempting heartbeats also for stopped canisters.
In this case a `CallMessage` is rejected and a `ResponseMessage` is generated by `rejectCallContext`. At least that's my educated guess.

marcin-dziadus: Maybe let's reject call contexts for heartbeats silently, i.e. make `rejectCallContext` conditional on the type of a call.

nomeata: Hmm, stopped canisters, I see. Stopped or stopping?

I think the cleanest in this case is to simply not run heartbeats for stopped (and stopping) canisters, rather than oddly ignoring stuff later.

I hope some later refactoring will clean this up a bit. Some things don't feel right.

I wonder if the spec formalism has the same or similar bugs?

marcin-dziadus: The spec is actually right - a canister needs to be running and not empty.

marcin-dziadus: There are some additional preconditions for starving and deleting, too. I updated the code accordingly.

src/IC/Ref.hs
```diff
@@ -654,7 +661,7 @@ rejectCallContext ctxt_id r =
 deleteCallContext :: ICM m => CallId -> m ()
 deleteCallContext ctxt_id =
   modifyCallContext ctxt_id $ \ctxt ->
-    if responded ctxt == Responded False
+    if (responded ctxt == Responded False && (not (isOriginatedFromHeartbeat ctxt)))
```
nomeata, line 664:
```suggestion
    if responded ctxt == Responded False && not (isOriginatedFromHeartbeat ctxt)
```
Also, why not `origin ctx /= FromHeartbeat`? Seems to say the same thing but easier to read and no need for a long-worded helper function?

(May need an `deriving Eq` for the `Origin` type)

src/IC/Ref.hs
```diff
@@ -943,7 +950,8 @@ icUninstallCode r = do
     -- reject all call open contexts of this canister
     gets (M.toList . call_contexts) >>= mapM_ (\(ctxt_id, ctxt) ->
         when (canister ctxt == canister_id && responded ctxt == Responded False) $ do
-            rejectCallContext ctxt_id (RC_CANISTER_REJECT, "Canister has been uninstalled")
+            when (not (isOriginatedFromHeartbeat ctxt)) $
+                rejectCallContext ctxt_id (RC_CANISTER_REJECT, "Canister has been uninstalled")
```
nomeata, line 954: This special casing feels odd. And I think it woudn’t be necessary if you set `reponded = True` when creating the call context.

But you might argue that it’s wrong to say that the call context has been responded to, because it hasn’t.

Suggestion:  Rename the `responded` field to `needs_response` in the Call Context type. Then it’s quite natural to start with `needs_response = False` for this call context, while normal calls begin with `needs_response = True` and turn to `needs_response = False` when a response has been generated. I think much will fall into place then.

src/IC/Ref.hs
```diff
@@ -1164,7 +1173,9 @@ setAllTimesTo ts = modify $
 runHeartbeat :: ICM m => CanisterId -> m ()
 runHeartbeat cid = do
   can <- getCanister cid
-  unless (idleSinceLastHeartbeat (last_action can)) $ do
+  is_empty   <- isCanisterEmpty cid
+  is_running <- isCanisterRunning cid
+  unless (idleSinceLastHeartbeat (last_action can) || is_empty || (not is_running)) $ do
```
nomeata, line 1178:
```suggestion
  unless (idleSinceLastHeartbeat (last_action can) || is_empty || not is_running) $ do
```

src/IC/Ref.hs
```diff
@@ -1125,8 +1134,8 @@ willReceiveResponse ic c = c `elem`
 nextStarved :: ICM m => m (Maybe CallId)
 nextStarved = gets $ \ic -> listToMaybe
   [ c
-  | (c, CallContext { responded = Responded False } ) <- M.toList (call_contexts ic)
-  , not $ willReceiveResponse ic c
+  | (c, ctx@CallContext { responded = Responded False } ) <- M.toList (call_contexts ic)
+  , not $ (willReceiveResponse ic c) || (isOriginatedFromHeartbeat ctx)
```
nomeata, line 1138:
```suggestion
  , not $ willReceiveResponse ic c || isOriginatedFromHeartbeat ctx
```
(Function application always binds more tightly than operators)

nomeata: Actually, if `responded = Responded True` always holds for Heartbeat Call Contexts, you shouldn’t need to change this code at all, which is a good sign.

src/IC/Ref.hs
nomeata: Actually, if responded = Responded True always holds for Heartbeat Call Contexts, you shouldn’t need to change this code at all, which is a good sign. (See comment below for why I think that would work)

src/IC/Ref.hs
```diff
@@ -257,12 +257,17 @@ canisterMustExist cid =
...
+    IsRunning -> return True
+    _         -> return False
 
 isCanisterEmpty :: ICM m => CanisterId -> m Bool
 isCanisterEmpty cid = isNothing . content <$> getCanister cid
 
+isOriginatedFromHeartbeat :: CallContext -> Bool
+isOriginatedFromHeartbeat ctx = case (origin ctx) of
+    FromHeartbeat -> True
+    _             -> False
```
nomeata, line 269: We might not need that, see comments below.  This comment just a reminder to omit this if that’s indeed true.

src/IC/Ref.hs
marcin-dziadus: That simplifies everything so much and makes the code consistent... Great idea, thanks!
However, we have to change the spec accordingly.

marcin-dziadus: The only problem with this refactoring is that: `responded` != `needs_response`, so we need to flip all conditions using that field and they are all over the place :) It's very mechanical though.

nomeata: Right. I guess `owes_response` might have avoided that :-D. So many ways to skin the cat

nomeata (approved):
I think I ran out of things to nitpick about.

src/IC/Ref.hs
marcin-dziadus: That's a creative way of solving the problem. I went for a crude one :D

marcin-dziadus:
> I think I ran out of things to nitpick about.

Thank you! I would've been nowhere close to this point without your help!

nomeata:
I hope it wasn’t too tedious for you. I think things will get better as you get more acquainted with the code base.

marcin-dziadus:
> I hope it wasn’t too tedious for you. I think things will get better as you get more acquainted with the code base.

No, it was very educational and I think it will pay off in the future. Thanks again!

basvandijk (approved):

nomeata:
BTW, I suggest to always use squash merge, and manually edit the commit message at the end, to have a useful git history. (At least until #33 will do that for us by taking a hopefully nicely written PR description.)

marcin-dziadus:
> > Did another pass. Overall good.
> > I wonder if deleting canisters in `simpleTestCase` is actually a win, given that it’s an extra 2-3s per test to wait. But 🤷🏻 .
> > When I worked on these things, and the replica already implemented the feature, I usually tested my ic-hs branch and latest replica. That was pretty easy when we were still using nix (just use `NIV_OVERRIDE`), but I have no idea how you would do that kind of testing now. Worth finding out and documenting, though!
> > I would be annoying to not test that now, then merge this, maybe even make a tagged release together with the spec, bump it in the `dfinity` repo, and _then_ notice that the `ic-ref-test` tests don’t work well together with the replica.
> 
> That's a perfect point! I would be annoyed, too.

Update: all tests passed.

PR #34, opened 2021-09-20
fix(#28): implement 64-bit stable memory support

### LINKED ISSUE #28: Implement 64-bit stable memory APIs
We would like to implement the new 64-bit stable memory APIs in the reference implementation so that we can also make a new release of the Interface spec and have end users start using the extended stable memory.

Looking into the code a bit I figured out that the main blocker right now is the `Memory` [module](https://github.com/dfinity-side-projects/winter/blob/e62722796e11760947a94285734e7bc29d72c389/src/Wasm/Runtime/Memory.hs) exposed by Winter which supports 32-bit memories and thus cannot be used to support the 64 bit stable memory APIs.

My idea on how this can be introduced in the least disruptive way is to provide a 64-bit Memory in Winter alongside the 32-bit one. Then the WinterMemory wrapper [module](https://github.com/dfinity/ic-hs/blob/c49a2f443d177b6486a1797ca646d8546ac7288c/src/IC/Wasm/WinterMemory.hs) can use this new 64-bit memory for either the old APIs (checking of course that the size is not bigger than allowed and casting 32 bit integers to 64 bit where necessary) or the new APIs. I think this would also be in line with future extensions of Winter (to support 64-bit wasm memories I assume you'd need a 64 bit memory implementation).

If the above sounds like a reasonable plan on a high level, then the next question would be how to best achieve this in practice. Looking at the code, it seems that we could maybe make [MemoryInst](https://github.com/dfinity-side-projects/winter/blob/e62722796e11760947a94285734e7bc29d72c389/src/Wasm/Runtime/Memory.hs#L50-L53) generic by the size of memory and save us from some ugly code duplication. This seems the most promising approach and one that should relatively minimize the amount of changes in general.

I'm curious to see what others think and open to other ideas.

roman-kashitsyn:
This change implements 64 bit stable memory API.

```
ic0.stable64_size : () -> (page_count : i64);                               // *
ic0.stable64_grow : (new_pages : i64) -> (old_page_count : i64);            // *
ic0.stable64_write : (offset : i64, src : i64, size : i64) -> ();           // *
ic0.stable64_read : (dst : i64, offset : i64, size : i64) -> ();            // *
```

This change introduces a new representation of stable memory:
we use size + history of writes instead of a flat byte buffer.
This results in a pretty straightforward code and allows us to handle
even huge heaps gracefully, which is crucial for testing interaction
between 32 and 64 bit APIs.

Fixes #28.


nomeata:
> The code uses exactly the same implementation for 32 and for 64 bit APIs.

I thought the spec says that the 32 bit variants will always `trap` (even if they could do something useful otherwise) if the stable memory is larger than 2^32 bytes. These checks probably need to be added to the system api implementation. At that point duplicating the four functions in src/IC/Canister/Imp.hs  might be cleaner than overloading, this way the code stays closer to the pseudo-code in the spec.

src/IC/Canister/Imp.hs
```diff
@@ -436,7 +447,31 @@ systemAPI esref =
     stable_read (dst, src, size) = do
       m <- gets stableMem
       i <- getsES esref inst
-      blob <- Mem.read m (fromIntegral src) size
+      blob <- Mem.read m (fromIntegral src) (fromIntegral size)
+      setBytes i (fromIntegral dst) blob
```
nomeata, line 451: `stable_read`, `stable_write`, `stable_size` should all have a predicate
```

      n <- Mem.size m
      when (n > 65536) $
        throwError "stable memory error: cannot use 32 bit API once stable memory is above 4GiB"
```
to match the specification, shouldn’t they?

nomeata (commented):
This needs a few tests in `ic-ref-test`, as well, right?

nomeata (approved):
Looks good now I'd say

roman-kashitsyn (commented):
> This needs a few tests in ic-ref-test, as well, right?

Yes, definitely. Though testing > 4GiB memories is tricky.
I can afford doing this as my machine has 64GiB of RAM:
```
    Growing stable memory beyond 4GiB         (24.28s)
    Using 32 bit API with large stable memory (76.52s)
```
The ic-ref process consumed 24GiB of memory at this point.

src/IC/Canister/Imp.hs
```diff
@@ -415,28 +421,66 @@ systemAPI esref =
...
+      n <- Mem.size m
+      when (n > 65536) $
+        throwError "stable memory error: cannot use 32 bit API once stable memory is above 4GiB"
+
     stable_size :: () -> HostM s Int32
     stable_size () = do
+      checkStableMemorySize
       m <- gets stableMem
-      Mem.size m
+      fromIntegral <$> Mem.size m
```
dsarlis, line 435: Well, you know what... I almost had this working when I forgot to add `<$>` which I don't even know what it does and I kept getting a weird error that there isn't a `fromIntegral` from the type I wanted to convert from. Can someone explain what this magic is doing?

src/IC/Test/Spec.hs
```diff
@@ -951,6 +958,57 @@ icTests = withAgentConfig $ testGroup "Interface Spec acceptance tests"
...
+
+    step "Growing stable memory beyond 4GiB"
+    call cid (replyData (i64tob (stable64Grow (int64 65535)))) >>= is "\x2\x0\x0\x0\x0\x0\x0\x0"
+    query cid (replyData (i64tob stable64Size)) >>= is "\x01\x00\x01\x00\x0\x0\x0\x0"
+
+    step "Using 32 bit API with large stable memory"
+    query' cid (ignore stableSize) >>= isReject [5]
+    query' cid (ignore $ stableGrow (int 1)) >>= isReject [5]
+    query' cid (stableWrite (int 0) "BAZ") >>= isReject [5]
+    query' cid (ignore $ stableRead (int 0) (int 3)) >>= isReject [5]
```
dsarlis, line 1010: Should we also do a write/read in 64 bit mode after we've grown the memory to make sure that works too?

dsarlis (approved):

src/IC/Canister/Imp.hs
dsarlis: Thanks, that helps explain it. To be fair, I understood from the error message that the problem is that I'm trying to apply `fromIntegral` directly to a higher order type which wouldn't work but I didn't know how to fix it. Anyway, TIL something.

src/IC/Test/Spec.hs
roman-kashitsyn: Good point!

It was somewhat slow with the flat memory representation (each new check was adding 30sec to the test runtime and 4GiB of process memory consumption), but the new representation makes adding tests a breeze.

Done!

roman-kashitsyn (commented):
PTAL, I've changed the stable memory representation.

This one actually worked the first time it compiled (I had a few crashes with MutableByteArray :() and the ic-ref-test runtime is very fast now.

nomeata:
Looks good! I might have used mutable vectors instead of ptr stuff. 

Kudos for writing a unit test suite!

PR #35, opened 2021-09-23
Fix build

nomeata:
looks like #28 included a warnings which broke the build.

And it looks like branch protection does block on the nix build. I’d
like to fix that, but can’t, for lack of permissions. Could someone with
admin rights please change the branch protection for `master` to block
on the nix build?


nomeata:
Hmm, I guess the nix build doesn’t even run for PRs. Gotta fix that…

marcin-dziadus (approved):

PR #37, opened 2021-09-24
Nix: Track haskell-upgrades

nomeata:
in #30 we tracked a PR against nixpkgs, which isn't very stable (e.g. if
it gets force-pushed to, it may break this repository). The changes we
are trying to include are still not in nixpkgs master, but they are
closer already, on the `haskell-upgrades` branch. So track that, until
it is merged into `master` (via https://github.com/NixOS/nixpkgs/pull/138596)

### DIFF
--- a/nix/sources.json
+++ b/nix/sources.json
@@ -39,16 +39,16 @@
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
     },
     "nixpkgs": {
-        "branch": "ghc-use-ghc8107-for-bootstrapping",
+        "branch": "haskell-updates",
         "builtin": true,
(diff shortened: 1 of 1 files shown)


cdepillabout:
Not sure if you've seen it, but there is a convenient site to tell when a PR you care about gets into master or the relevant channels: https://nixpk.gs/pr-tracker.html?pr=138596

nomeata:
(This is a draft until https://github.com/nomeata/ic-hs/runs/3698287099?check_suite_focus=true is green. Recompiling all the musl/static stuff always takes a while)

marcin-dziadus (approved):

marcin-dziadus:
> (This is a draft until https://github.com/nomeata/ic-hs/runs/3698287099?check_suite_focus=true is green. Recompiling all the musl/static stuff always takes a while)

It's finally green.

PR #40, opened 2021-09-30
Keep track of the number of instructions executed so far.

marcin-dziadus:


src/IC/Ref.hs
```diff
@@ -136,6 +136,7 @@ data CanState = CanState
   , time :: Timestamp
   , cycle_balance :: Natural
   , certified_data :: Blob
+  , executed_instructions :: Natural
```
nomeata, line 139: Hmm, this looks awfully complicated. Isn't the performance counter a per-message concept? If so, I would expect it doesn't affect this module at all, and can be completely be contained in `IC.Canister.Imp`. or am I missing something?

src/IC/Test/Spec.hs
```diff
@@ -971,6 +972,12 @@ icTests = withAgentConfig $ testGroup "Interface Spec acceptance tests"
       query cid (replyData getGlobal) >>= as2Word64 >>= bothSame
     ]
 
+  , testGroup "performance_counter"
+    [ testCase "simple" $ do
+        cid <- install (setGlobal (i64tob performanceCounter))
+        query cid (replyData getGlobal) >>= asWord64 >>= is 16
```
nomeata, line 978: I don't expect you can assert specific values here. Not until all aspects of cycle accounting are rigidly specified in the interface spec, and we might not even _want_ that. Yes, that means there may not be much we can test here - just like with cycle accounting in general.

src/IC/Wasm/Winter.hs
```diff
@@ -114,18 +115,26 @@ exportedFunctions wasm_mod =
...
+invokeExport :: Instance s -> FuncName -> [W.Value] -> HostM s Natural
 invokeExport (mods', ref) method args = do
   let inst = mods' IM.! ref
-  withExceptT show $
-    W.invokeByName mods' inst (T.pack method) args
-
-invokeTable :: Instance s -> Int32 -> [W.Value] -> HostM s [W.Value]
+  withExceptT show $ W.runEvalT $ do
+    _ <- W.invokeByName mods' inst (T.pack method) args
+    ic <- W.invoke mods' inst W.getInstructionCount []
```
nomeata, line 123: It seems that winter is providing some instructon counter? But unless that counter happens to behave _exactly_ like the instruction counting in the replica, this is not a useful number. And thus I'd rather use something blatantly different (e.g. always 0, or bump by one in each system call, or something like that - the simplest thing we can get away with)

src/IC/Test/Spec.hs
marcin-dziadus: Sorry, it was just my experiment. I wanted to see what's returned here and forgot to remove this test. :)

src/IC/Ref.hs
marcin-dziadus: Yes, it's indeed a per-message concept and that's why this implementation is incorrect (not spec-compliant).
This counter keeps track of the number of instruction executed "since the beginning of time" and it's updated after each message is processed, so it's wrong for a few reason...

src/IC/Wasm/Winter.hs
marcin-dziadus: Yes, we can fetch from Winter the number of steps the WASM interpreter took.
However, we can get this number only after the execution is finished which is a serious limitation. (maybe I am mistaken here)

marcin-dziadus:
@nomeata I think I'll keep [#41](https://github.com/dfinity/ic-hs/pull/41) for now and close this draft PR. It's a playground PR, rather than a promising attempt, after all. Does that make sense?

nomeata:
Yes, I think we can’t do more than #41, nor do we actually need to do more.

PR #41, opened 2021-09-30
Create a stub of performance_counter

marcin-dziadus:
The stub will be replaced with a proper implementation once its semantics gets clarified.

### DIFF
--- a/src/IC/Test/Spec.hs
+++ b/src/IC/Test/Spec.hs
@@ -1044,12 +1044,6 @@ icTests = withAgentConfig $ testGroup "Interface Spec acceptance tests"
       query cid (replyData getGlobal) >>= as2Word64 >>= bothSame
     ]
 
-  , testGroup "performance counter" $
-    [ testCase "call once" $ do
-        cid <- install noop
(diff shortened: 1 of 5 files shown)


basvandijk (approved):

src/IC/Test/Spec.hs
```diff
@@ -1043,6 +1044,12 @@ icTests = withAgentConfig $ testGroup "Interface Spec acceptance tests"
       query cid (replyData getGlobal) >>= as2Word64 >>= bothSame
     ]
 
+  , testGroup "performance counter" $
+    [ testCase "call once" $ do
+        cid <- install noop
```
nomeata, line 1049: Remember that rhe test suite is _not_ just a test for any single implementation, but should only test whats in the spec. So this should not assert that it's 0, but merely that a call can be made.

marcin-dziadus: Ah, that's a very good point. It would almost certainly fail for any other implementation. Thanks! :)

nomeata (approved):

Issue #33, opened 2021-09-20
Enable mergify for this repo


nomeata:
[Mergify](mergify.io/) is a great tool that leaves merging to this bot, meaning less maual work for us.

One great feature is that it supports squash merges that take the _PR description_ as the commit message. So instead of a commit message like https://github.com/dfinity/ic-hs/commit/fb558eb58aed501a6b46768baf0a50ece6bedeaa we’d get something nicely curated like the PR description of https://github.com/dfinity/ic-hs/pull/30. You can achieve the same manually (as [explained here](https://github.com/dfinity/ic-hs/pull/30#issuecomment-922434754)), and that’s how I have been merging PRs in this repo so far, but it is tedious work, so won’t always work.

There is more that we can do with mergify, e.g. auto-merge dependency bumps.

We already have mergify enabled for the sister repository `dfinity/motoko`, so I see no problem doing it here too. Someone just has to enable it in https://github.com/organizations/dfinity/settings/installations/12369523, and I will prepare a PR with the configuration.

@basvandijk, are you still doing this kind of infra work? Or who should I nudge?

marcin-dziadus:
@nomeata Mergify should be already enabled!

nomeata:
Oh, indeed? Let’s try

marcin-dziadus:
I think we can close this issue. Thank you for your contribution!

PR #42, opened 2021-10-01
Configure mergify

nomeata:
this adds a mergify configuration to the repo, fixing #33.

TL;DR: Label a PR with `automerge-squash` and it will be merged once
approved an green. Also, mergify will update out-of-date branches as
`master` changes.

It is based on the `.mergify.yaml` from Motoko, but omitting the “fancy”
features (e.g. approving changelog changes and dependency bumps). We can
add them when and if we add auto dependency bumping, but we don’t have
that many dependencies, so maybe not needed.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,23 @@
+pull_request_rules:
+  - name: Automatic merge (squash)
+    conditions:
+      - "#approved-reviews-by>=1"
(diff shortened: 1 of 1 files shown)


nomeata:
Ah, indeed mergify reacts. Good sign!

marcin-dziadus (approved):

PR #49, opened 2021-10-14
Reply successfully to pre-flight requests

nmattia:
Without these cors settings, `ic-ref` does not respond to `OPTIONS`
requests, which are made by most browsers. These changes make `ic-ref`
behave more like the `replica`.

I don't think the spec mentions this.

@nomeata @roman-kashitsyn this is a follow up to https://github.com/dfinity/internet-identity/pull/425.

---

To check this, start `ic-ref` and issue either a curl preflight-y check:

```
$ curl -H Origin: http://localhost:8080 -H Access-Control-Request-Method: GET -H Access-Control-Request-Headers: X-Requested-With -X OPTIONS --verbose http://127.0.0.1:8001/api/v2/canister/fterm-bydaq-aaaaa-aaaaa-c/query
```

or try to call any canister from a browser:

**before**

Screenshot 2021-10-14 at 11 41 17

**after**

Screenshot 2021-10-14 at 11 41 29


nomeata:
You probably have to update the freeze file

PR #54, opened 2021-10-29
Support 128-bit system API for cycles

marcin-dziadus:


default.nix
```diff
@@ -11,8 +11,8 @@ let universal-canister = (naersk.buildPackage rec {
     name = "universal-canister";
     src = subpath ./universal-canister;
     root = ./universal-canister;
-    CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_LINKER = "${nixpkgs.llvmPackages_11.lld}/bin/lld";
-    RUSTFLAGS = "-C link-arg=-s"; # much smaller wasm
+    CARGO_TARGET_WASM32_UNKNOWN_UNKNOWN_LINKER = "${nixpkgs.llvmPackages_12.lld}/bin/lld";
+    RUSTFLAGS = "-C link-arg=-s -C target-feature=+multivalue"; # much smaller wasm
```
nomeata, line 15: I’m surprised this work, given https://github.com/rust-lang/rust/issues/73755, but 🤷🏻  :-)

src/IC/Canister/Imp.hs
```diff
@@ -335,27 +342,52 @@ systemAPI esref =
         Stopping -> 2
         Stopped -> 3
 
+    splitBitsIntoHalves :: Natural -> (Word64, Word64)
+    splitBitsIntoHalves n = (fromIntegral $ highBits n, fromIntegral $ lowBits n)
+        where highBits = flip shiftR 64
+              lowBits = (.&.) (1 `shiftL` 64 - 1)
```
nomeata, line 348: I often find it easier to just write literals, but it's a matter of taste (needs `{-# LANGUAGE NumericUnderscores #-}` for the `_`)
```suggestion
              lowBits = (0xFFFFFFFF_FFFFFFFF .&.)
```

src/IC/Canister/Imp.hs
```diff
@@ -335,27 +342,52 @@ systemAPI esref =
...
+
+    combineBitHalves :: (Word64, Word64) -> Natural
+    combineBitHalves (high, low) = fromIntegral high `shiftL` 64 + fromIntegral low
```
nomeata, line 351: Doesn’t make a differnce, but when dealing with bit-wise stuff I find it cleaner to use bit operations rather than numeric operations.
```suggestion
    combineBitHalves (high, low) = fromIntegral high `shiftL` 64 .|. fromIntegral low
```

src/IC/Test/Agent.hs
```diff
@@ -839,3 +843,6 @@ textual = T.unpack . prettyPrincipal . Principal
 shorten :: Int -> String -> String
 shorten n s = a ++ (if null b then "" else "…")
   where (a,b) = splitAt n s
+
+toI128 :: (Word64, Word64) -> Natural
+toI128 (high, low) = fromIntegral high `shiftL` 64 + fromIntegral low
```
nomeata, line 848:
```suggestion
toI128 (high, low) = fromIntegral high `shiftL` 64 .|. fromIntegral low
```

src/IC/Test/Spec.hs
```diff
@@ -1576,7 +1583,19 @@ icTests = withAgentConfig $ testGroup "Interface Spec acceptance tests"
           ic_install (ic00via cid) (enum #install) cid2 universal_wasm (run noop)
           return cid2
     in
-    [ testGroup "can use balance API" $
+    [ testGroup "cycles API - backward compatibility" $
+        [ simpleTestCase "canister_cycle_balance = canister_cycle_balance128 for numbers fitting in 64 bits" $ \cid -> do
+          a <- queryBalance cid
+          b <- queryBalance128 cid
+          bothSame ((0,a),b)
+        , testCase "lagacy API traps when a result is too big" $ do
```
nomeata, line 1591:
```suggestion
        , testCase "legacy API traps when a result is too big" $ do
```

nomeata (approved):
Nice work

src/IC/Canister/Imp.hs
marcin-dziadus: I like it better, too. Thanks!

marcin-dziadus:
Thanks for the review, @nomeata!

default.nix
marcin-dziadus: I was quite surprised too, when I found it working. I gave it a go and I did not regret. :)

However, there is one last problem. Even though the produced WASM is correct, the compiler spits out following warnings: "warning: `extern` block uses type `(u64, u64)`, which is not FFI-safe". We can mitigate this at a price of obfuscating types by using a tuple struct, i.e. `Pair(u64, u64)`. More radical option is to disable the troublesome check. What do you think?

nomeata: For `ic-ref-test`, I have no strong opinions. If using a tuple struct works, go head – what matters is what happens on the Wasm level.

ielashi (approved):
LGTM. cc @AlexandraZapuc

PR #55, opened 2021-10-30
[VER-1123] Implement metadata access

nomeata:
this is a reimplementation of @jwiegley’s #53; his PR gave me many ideas
that I felt would be very tedious and impolite to convey through prose
review, so I just wrote down how I would do it. I hope this achieves the
level of declaratively that we aim for (by moving plumbing code into
other modules, for example, and by having more precise types.)

In addition to #53, this performs more module validation, and adds
further tests, and fixes a bug where deleted canisters still show up in
the state tree (probably worth refactoring to avoid such bugs, but not
in this PR).

### DIFF
--- a/ic-hs.cabal
+++ b/ic-hs.cabal
@@ -336,6 +336,7 @@ executable ic-ref-test
   other-modules: IC.Test.Universal
   other-modules: IC.Types
   other-modules: IC.Version
+  other-modules: IC.Utils
   other-modules: SourceId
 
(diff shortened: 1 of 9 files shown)


jwiegley (approved):

nomeata:
@marcin-dziadus, do you want to do a review too? If only to familiarize yourself with the code? Feel free to merge when and if you are happy.

PR #57, opened 2021-11-11
Update system API for 128 bit cycles tests

AlexandraZapuc:
Following up on our discussion regarding the multivalue target feature support, we have decided to update the experimental System Api and not rely on this feature anymore.

This MR updates the tests according to the Public Spec.

### DIFF
--- a/src/IC/Canister/Imp.hs
+++ b/src/IC/Canister/Imp.hs
@@ -30,7 +30,7 @@ import qualified Data.Text as T
 import qualified Data.ByteString.Lazy as BS
 import qualified Data.ByteString.Lazy.Char8 as BSC
 import qualified Data.ByteString.Lazy.UTF8 as BSU
-import qualified Data.Binary as B
+import qualified Data.Binary.Put as Put
 import Control.Monad.Primitive
 import Control.Monad.ST
 import Control.Monad.Except
@@ -364,17 +364,18 @@ systemAPI esref =
     lowBits = fromIntegral . (0xFFFFFFFF_FFFFFFFF .&.)
 
     to128le :: Natural -> BS.ByteString
-    to128le n = BS.reverse $ BS.append (B.encode (highBits n)) (B.encode (lowBits n))
+    to128le n = Put.runPut $ do
+      Put.putWord64le (lowBits n)
+      Put.putWord64le (highBits n)
 
     combineBitHalves :: (Word64, Word64) -> Natural
     combineBitHalves (high, low) = fromIntegral high `shiftL` 64 .|. fromIntegral low
 
     low64BitsOrErr :: Natural -> HostM s Word64
-    low64BitsOrErr n = if highBits n > 0
-      then
+    low64BitsOrErr n = do
+      unless (highBits n == 0) $
         throwError $ "The number of cycles does not fit in 64 bits: " ++ show n
-      else
-        return $ fromIntegral $ lowBits n
+      return $ fromIntegral $ lowBits n
 
     msg_cycles_refunded :: () -> HostM s Word64
     msg_cycles_refunded () =  getRefunded esref >>= low64BitsOrErr
--- a/src/IC/Canister/Imp.hs
+++ b/src/IC/Canister/Imp.hs
@@ -30,6 +30,7 @@ import qualified Data.Text as T
 import qualified Data.ByteString.Lazy as BS
 import qualified Data.ByteString.Lazy.Char8 as BSC
 import qualified Data.ByteString.Lazy.UTF8 as BSU
+import qualified Data.Binary as B
 import Control.Monad.Primitive
 import Control.Monad.ST
 import Control.Monad.Except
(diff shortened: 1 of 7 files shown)


nomeata:
Thanks, that reduces my worries!

I wonder if its worth simplifying the universal canister interface, undo the addition of `I128` to the interface, and use two `I64` for the arguments, and `Blob` for the results. This way the universal canister does less and is closer to the system API. WDYT, @marcin-dziadus ?

marcin-dziadus:
> Thanks, that reduces my worries!
> 
> I wonder if its worth simplifying the universal canister interface, undo the addition of `I128` to the interface, and use two `I64` for the arguments, and `Blob` for the results. This way the universal canister does less and is closer to the system API. WDYT, @marcin-dziadus ?

Sorry for the late response - I was on vacation for a few days.
@nomeata Yes, it's a very good point! I'll take care of it.

marcin-dziadus:
@nomeata I'm done with the refactoring, could you please take a look?

marcin-dziadus (approved):

src/IC/Canister/Imp.hs
```diff
@@ -343,52 +357,61 @@ systemAPI esref =
...
-        where highBits = flip shiftR 64
-              lowBits = (0xFFFFFFFF_FFFFFFFF .&.)
+    highBits :: Natural -> Word64
+    highBits = fromIntegral . (flip shiftR 64)
+
+    lowBits :: Natural -> Word64
+    lowBits = fromIntegral . (0xFFFFFFFF_FFFFFFFF .&.)
+
+    to128le :: Natural -> BS.ByteString
+    to128le n = BS.reverse $ BS.append (B.encode (highBits n)) (B.encode (lowBits n))
```
nomeata: I suggest to use `Data.ByteString.Builder` here, one dependency less, and its what we use elsewhere in ic-hs and haskell-candid.

src/IC/Canister/Imp.hs
```diff
@@ -343,52 +357,61 @@ systemAPI esref =
...
 
-    low64BitsOrErr :: (Word64, Word64) -> HostM s Word64
-    low64BitsOrErr (0, low) = return low
-    low64BitsOrErr (high, low) = throwError $ "The number of cycles does not fit in 64 bits: " ++ show (combineBitHalves (high, low))
+    low64BitsOrErr :: Natural -> HostM s Word64
+    low64BitsOrErr n = if highBits n > 0
+      then
+        throwError $ "The number of cycles does not fit in 64 bits: " ++ show n
+      else
+        return $ fromIntegral $ lowBits n
```
nomeata: Purely syntatic suggestion, feel free to ignore. But guards usually look better than `if .. then ... else ...`, especially if the full RHS of a function is the if-then-else.

Alternative, a nice way to handle this (since `throw` actuallys throws) is to use `when` or `unless` for such checks, and then just continue below without indentation. This works well in imperative code like this.

nomeata:
About the failing check: A nne nix release nieeds an optio to  be set; have a look at https://github.com/dfinity/motoko/pull/2880 for how to fix this.

src/IC/Canister/Imp.hs
nomeata: Nevermind, looks like I was looking at an older commit here

src/IC/Canister/Imp.hs
marcin-dziadus: I think you looked at an old commit here, too :)
I rewrote it in a simpler and more readable way, but did not use `Builder`. Is it okay?

marcin-dziadus:
> About the failing check: A nne nix release nieeds an optio to be set; have a look at [dfinity/motoko#2880](https://github.com/dfinity/motoko/pull/2880) for how to fix this.

Thanks a lot!

src/IC/Canister/Imp.hs
nomeata: You switched to `Put`, which is from the `binary` package, while elsewhere we use https://hackage.haskell.org/package/bytestring-0.11.1.0/docs/Data-ByteString-Builder.html from the `bytestring` package for the same purpose (binary encodings). But its really no big deal, so feel free to merge as is.

marcin-dziadus: Why not be consistent if it comes at a low cost? Thanks, done.

nomeata:
@marcin-dziadus, guess we missed the chance to use `ic-ref-test` to catch bugs in production: It doesn't actually test if cycle balances above 2^64 work, right? If we did, this might have prevented https://forum.dfinity.org/t/panicked-at-failed-to-serialize-message-errorimpl-code-message-the-number-cant-be-stored-in-cbor-offset-0-build-rs-canister-sandbox-common-src-transport-rs55/10729?u=nomeata…

nomeata:
Hmm, nevermind, there is a test here:
```
          let large = 2^(65::Int)
          ic_top_up ic00 cid large
```
so now I am curious how that relates to the bug I reported at the forum.

Is `ic-ref-test` still run against the replica as part of CI?

marcin-dziadus:
Yes, it is, but we haven't released a new version for a while.

nomeata:
So the replica ships with features that don't have an spec acceptance test in CI yet? I guess we can't have everything

marcin-dziadus:
We should optimise the process. I will raise this issue.

PR #58, opened 2021-11-19
Set lower bound on aeson

nmattia:
The module `IC.Crypto.WebAuthn` uses `Aeson.parseFail` which was only
introduced in
[1.4.7](https://github.com/haskell/aeson/blob/master/changelog.md#1470)


nomeata (approved):
I was lazy with package bounds so far, given that the few users use nix or lockfiles. But it's good to put in whatever knowledge we gain about versions.

nomeata:
Hmm, I need to fix CI (need to set a nix option, did the same before in Motoko), but not via the phone.

PR #61, opened 2021-11-30
Include `canister_ranges` in the subnet information

marcin-dziadus:


src/IC/CBOR/Utils.hs
```diff
@@ -7,3 +7,6 @@ import IC.Types
 
 encodePrincipalList :: [EntityId] -> Blob
 encodePrincipalList entities = BS.toLazyByteString $ encode $ GList $ map (GBlob . rawEntityId) entities
+
+encodeCanisterRangeList :: [CanisterRange] -> Blob
+encodeCanisterRangeList ranges = BS.toLazyByteString $ encode $ GList $ map (\r -> GList $ map (GBlob . rawEntityId) [fst r, snd r]) ranges
```
nomeata, line 12:
```suggestion
encodeCanisterRangeList ranges = BS.toLazyByteString $ encode $ GList $ map (\(l,u) -> GList $ map (GBlob . rawEntityId) [l, u]) ranges
```

src/IC/Ref.hs
```diff
@@ -572,12 +575,13 @@ getPrunedCertificate time paths = do
     let cert_tree = prune full_tree (["time"] : paths)
     sk1 <- gets secretRootKey
     sk2 <- gets secretSubnetKey
-    return $ signCertificate time sk1 (Just (fake_subnet_id, sk2)) cert_tree
+    cs <- M.keys <$> gets canisters
+    return $ signCertificate time sk1 (Just (fake_subnet_id, sk2, (minimum cs, maximum cs))) cert_tree
```
nomeata, line 579: This will crash when there are no canisters, right?

marcin-dziadus: Very good point, thanks!

src/IC/Ref.hs
```diff
@@ -575,12 +575,15 @@ getPrunedCertificate time paths = do
...
     sk2 <- gets secretSubnetKey
-    cs <- M.keys <$> gets canisters
-    return $ signCertificate time sk1 (Just (fake_subnet_id, sk2, (minimum cs, maximum cs))) cert_tree
+    cans <- M.keys <$> gets canisters
+    let r = (\cs -> (minimum cs, maximum cs)) <$> list_to_maybe cans
+    return $ signCertificate time sk1 (Just (fake_subnet_id, sk2, r)) cert_tree
   where
     fake_subnet_id = EntityId "\x01"
+    list_to_maybe [] = Nothing
+    list_to_maybe x = Just x
```
nomeata, line 584: Hmm, this is all a bit roundabout, and more complex than it has to be.

Here is a simpler way: given the code in https://github.com/dfinity/ic-hs/blob/f19a1621f5bc3609e6f003fdcc0eb9cab77ab224/src/IC/Id/Fresh.hs you can just use a static range of canister IDs (0x000000000000000000 to 0xffffffffffffffffff if I am not mistaken, but it may be cleaner to just calculate the them based using `minBound` and `maxBound`.

So I'd define the tuple of bounds statically, and to keep the logic next to each other, define it in `IC.Id.Fresh`. WDYT?

marcin-dziadus: @nomeata I merged the PR prematurely... Anyway, I will send another PR which simplifies that code according to your suggestion.

marcin-dziadus: On the other hand, using `[(minBound, maxBound)]` is semantically different compared to using an empty list.
The former means that any ID belongs to the subnet, the later that no ID belongs to it.

nomeata: I meant use `minBound :: Word64` but then turn into a principal as done in IC.Id.Fresh, so it becomes the range of canister IDs used by ic-ref (which is _almost_ all possible principals, but not all :-))

PR #62, opened 2021-12-01
Simplify creation of `canister_ranges`

marcin-dziadus:
This PR is a follow-up to #61

### DIFF
--- a/src/IC/Id/Fresh.hs
+++ b/src/IC/Id/Fresh.hs
@@ -11,5 +11,11 @@ freshId :: [EntityId] -> EntityId
 freshId ids =
     head $
     filter (`notElem` ids) $
-    map (EntityId . mkOpaqueId . toLazyByteString . word64LE)
+    map wordToId
     [1024::Word64 ..]
+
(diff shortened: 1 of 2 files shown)


src/IC/Id/Fresh.hs
```diff
@@ -11,5 +11,11 @@ freshId :: [EntityId] -> EntityId
 freshId ids =
     head $
     filter (`notElem` ids) $
-    map (EntityId . mkOpaqueId . toLazyByteString . word64LE)
+    map wordToId
     [1024::Word64 ..]
+
+icCanisterIdRange :: (EntityId, EntityId)
+icCanisterIdRange = (wordToId minBound, wordToId maxBound)
```
marcin-dziadus, line 18: @nomeata Shouldn't we use `1024` here?

nomeata: Either is fine, I didn't know think it makes a difference

nomeata (approved):
Yay, less code, more happy!

PR #64, opened 2021-12-05
Bump `nixpkgs` to 21.11

nomeata:


ggreif (approved):

nomeata:
It seems to work on Linux, but the distribution build on MacOS doesn't anymore, maybe missing some system libraries. Not sure how well I can solve this with blind trial and error

ggreif:
> with blind trial and error

Heh, I followed my intuition (and had a local build) :-)

PR #69, opened 2022-01-06
ic-ref-run: Execute heartbeats before any submitted message

### LINKED ISSUE #68: ic-ref-run has no heartbeat functionality
drun triggers the heartbeat, but ic-ref-run appears not to, though ic-ref supports it. This means we can only test heartbeat functionality on drun for Motoko.

nomeata:
fixes #68.

I did not check if this matches `drun`’s semantics (heartbeat before
each message? after each message? time-based). But shouldn’t matter.


nomeata:
@crusso, this should do. You can test it on your PR by temporarily switching the `ic-hs` pin to this branch (`niv update ic-hs -b joachim/ic-ref-run-heartbeat`).

crusso:
@nomeata sorry - completely overlooked this - still want me to test it?

src/ic-ref-run.hs
```diff
@@ -93,6 +93,7 @@ submitAndRun :: CallRequest -> DRun ()
 submitAndRun r = do
     lift $ printCallRequest r
     rid <- lift mkRequestId
+    processHeartbeats
```
crusso, line 96: Not that's concise

nomeata:
Yes, testing would be good :-)

crusso:
Where does the time go? Doing a quick test now...

crusso:
Not so quick after all - the run-drun/heartbeat.mo test just hangs for me after outputting:
 ```
 [nix-shell:~/motoko/test/run-drun]$ more _out/heartbeat.ic-ref-run 
→ update create_canister(record {settings = null})
← replied: (record {hymijyo = principal "cvccv-qqaaq-aaaaa-aaaaa-c"})
→ update install_code(record {arg = blob ""; kca_xin = blob "\00asm\01\00\00\00\
0…
← replied: ()
→ update go()
```

crusso:
c.f. https://github.com/dfinity/motoko/pull/3071

crusso:
On ic-ref, the heartbeat seems to get invoked only once AFAICT (verified by inserting a debug print in heartbeat.mo's heartbeat function).

crusso:
@nomeata any idea what's going wrong?

crusso:
Good point. I've asked @stefan-at-dfinity if drun could provide that, but I can just easily do as you propose and add some dummy ingress messages to drive the heartbeat.  I think my test just exits after n heartbeats at the moment, so is hanging on ic-ref.

crusso:
User relying on this in wild:

https://github.com/enzoh/motoko-qr/blob/master/app/Demo.mo

roman-kashitsyn:
> I would say that what we want is a drun command, like query and call, called heartbeat <cid> which invokes the heartbeat of a specific canister once. And then use that to script our tests.

I'm not that's an appropriate interface.

I see `drun` is a simple language to script calls compliant with the public spec. In theory, you could implement DRUN by talking to the mainnet (maybe except `debug.print`s).

There is no call to the IC to execute a heartbeat. If we want to make it controlled, I suggest extending `drun` to support `tick` command that submits an empty block and waits for the state height to increase. This will be pretty straightforward to implement (unlike `heartbeat <cid>`, which needs a new tunnel through component interfaces).

ulan:
Giving `drun` more control over execution of heartbeats sgtm in general.

My main concern is that `drun` is a wrapper around our main implementation. Supporting callable heartbeats would require plumbing in the production code of the execution environment and would add more the complexity there. So the main question is whether the added complexity is worth it given that there is a simple workaround of using  "some dummy ingress messages to drive the heartbeat."

nomeata:
Fully agree: it's a trade off whenever non-production features enter production code, and if the `tick` command is a good compromise, then that's of course fine. I'm just saying that it'd be a compromise, and not the ideal.

mraszyk:
@nomeata I wanted to resolve conflicts and this made this PR closed. I've opened #114 as a follow-up.

PR #71, opened 2022-02-01
Enable canisters to send HTTP requests

marcin-dziadus:


nomeata:
Hmm, I wonder if it goes against the spirit of a reference implementation to actually talk to the internet. This means that it’s behaviour becomes non-deterministic, and a bunch of external factors influence it.

Have you considered just mocking the component? I.e. only supporting a bunch of fixed, hard-coded URLs with fixed behaviors?

This might make it actually easier to, for example, create interesting corner cases to test for example a canister’s response normalization function.

src/IC/Ref.hs
```diff
@@ -860,6 +867,26 @@ invokeManagementCanister caller ctxt_id (Public method_name arg) =
...
 invokeManagementCanister _ _ Heartbeat = error "heartbeat invoked on management canister"
 
+sendHttpRequest :: T.Text -> IO (C.Response BS.ByteString)
+sendHttpRequest url = do
+  m <- C.newManager C.defaultManagerSettings
+  initReq <- C.parseRequest (T.unpack url)
+  let req = initReq {
+      C.method = "GET"
+  }
+  C.httpLbs req m
```
nomeata, line 877: This could be moved to a “plumbing” module, rather than a “reference” module (as per the distinction described at https://github.com/dfinity/ic-hs/blob/master/README.md#ic-ref-a-ic-reference-implementation)

marcin-dziadus:
> Hmm, I wonder if it goes against the spirit of a reference implementation to actually talk to the internet. This means that it’s behaviour becomes non-deterministic, and a bunch of external factors influence it.
> 
> Have you considered just mocking the component? I.e. only supporting a bunch of fixed, hard-coded URLs with fixed behaviors?
> 
> This might make it actually easier to, for example, create interesting corner cases to test for example a canister’s response normalization function.

Honestly, exactly the same concerns crossed my mind when I was writing this code. What do you precisely mean by "mocking the component"? Could you please elaborate? In particular, how can we do it and avoid having implementation-specific behaviours?

nomeata:
I don’t fully know yet either. The nice thing about the Interface Spec and `ic-ref` so far is that it’s completely passive: You poke it from the outside, but it’s self-contained. This now changes.

So what roles does `ic-ref` have? 

* As code that explains the spec. For that, it doesn't matter what happens behind this call.

* As _runnable_ code that you can test canisters with (useful for Motoko and internet-identity). Here it might be sufficient if it just “made up” the HTTP responses.
 
  For example, it could pretend that the internet only consisted of one page, `example.com` (all other URLs are treated as non-existant), and there are only a handful of pages at that URL, one for a few possible interesting behaviors (text, binary data, 404, varying data that can be normalized maybe).

  This would be sufficient to test CDKs or canisters that interact with HTTP in a generic way.

* A full simulator. This would have to talk to the real internet.

* A middle ground, where responses are mocked, but not hard-coded but read from some JSON file or so, so that canister developers can write their test cases.

I’d lean towards the second, or even the third.  I am not sure how useful ic-ref is as a full simulator (it doesn't do cycles, for example).

The same question will come up with bitcoin integration, of course.

marcin-dziadus:
Thank you for the detailed response, @nomeata!
One thing is unclear to me: if we return a made-up response or mock a HTTP client in `ic-ref`, how would we keep `ic-ref-test` universal? Of course, we still can verify some nontrivial properties of the implementation, e.g. that the `transform` function is called and so on, but it would be impossible to make any particular assumptions about the returned HTTP response. 

In this PR I addressed the problem through starting a dummy HTTP server to ensure that all implementations get exactly the same response.

nomeata:
> In this PR I addressed the problem through starting a dummy HTTP server to ensure that all implementations get exactly the same response.

Hmm, maybe that is the way to go then. So essentially there is a “mocked” HTTP server with predfined behaviour, and it comes with `ic-ref-test`. Yeah, that’s actually a good way forward, thanks!

marcin-dziadus:
> > In this PR I addressed the problem through starting a dummy HTTP server to ensure that all implementations get exactly the same response.
> 
> Hmm, maybe that is the way to go then. So essentially there is a “mocked” HTTP server with predfined behaviour, and it comes with `ic-ref-test`. Yeah, that’s actually a good way forward, thanks!

One problem remains - how to test logic around the `transform` function. It's possible only if a canister exports a function with a name corresponding to the one in a request (and a proper signature). How can we ensure that? Maybe we shouldn't cover this part at all...

nomeata:
Not sure what you mean. The transform function is a reference to a query method, identified by principal and method name. Surely `ic-ref` has no problem executing that, just like any other query method?

marcin-dziadus:
> Not sure what you mean. The transform function is a reference to a query method, identified by principal and method name. Surely `ic-ref` has no problem executing that, just like any other query method?

Agreed, `ic-ref` has no problem executing it, but the function has to exist, so if I want to verify in `ic-ref-tests` that `transform` works as intended, I need to modify the universal canister (here and for the replica), right?
Sorry, if I am missing something obvious...

nomeata:
> I need to modify the universal canister (here and for the replica), right?

Ah, yes! The universal canister isn’t universal enough :-)

src/IC/Ref.hs
marcin-dziadus: I couldn't find a good place for it... Does that make sense to put it into a separate module?

nomeata: Yes, modules are cheap :-)

Maybe `IC.Ref.IO` for when`IC.Ref` has side effects?

src/ic-ref-test.hs
```diff
@@ -37,6 +37,5 @@ main = do
         , antXMLRunner `composeReporters` htmlRunner `composeReporters` consoleTestReporter
         ]
       ]
-
-startEchoServer = forkIO $ run 8003 app
-  where app _ f = f $ responseLBS status200 [(hContentType, "text/plain")] "Hello world!"
\ No newline at end of file
+    startServer = forkIO $ run 8003 app
```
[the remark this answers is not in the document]
nomeata: It helps a bit, and can be generalized to auto-chosen ports when it actually becomes a problem

marcin-dziadus:
@nomeata Could you please help me with fixing Nix checks? I could not reproduce problems locally which makes me quite helpless...

ggreif:
The `linux` build is surely dying due to low memory (OOM-killer). The MacOS build id dying due to the missing
```
  other-modules: IC.Ref.IO
```
In line 474 of `ic-hs.cabal`.

marcin-dziadus:
> @marcin-dziadus:
> 
> The `linux` build is surely dying due to low memory (OOM-killer). The MacOS build id dying due to the missing
> 
> ```
>   other-modules: IC.Ref.IO
> ```
> 
> In line 474 of `ic-hs.cabal`.

Thank you very much, Gabor!

ggreif:
Here are more errors:
```
[30 of 49] Compiling IC.Ref.IO        ( src/IC/Ref/IO.hs, dist/build/unit-test/unit-test-tmp/IC/Ref/IO.o, dist/build/unit-test/unit-test-tmp/IC/Ref/IO.dyn_o

src/IC/Ref/IO.hs:10:1: error:
    Could not load module ‘Network.HTTP.Client’
    It is a member of the hidden package ‘http-client-0.6.4.1’.
    Perhaps you need to add ‘http-client’ to the build-depends in your .cabal file.
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
   |
10 | import qualified Network.HTTP.Client as C
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

src/IC/Ref/IO.hs:11:1: error:
    Could not load module ‘Network.HTTP.Types.Status’
    It is a member of the hidden package ‘http-types-0.12.3’.
    Perhaps you need to add ‘http-types’ to the build-depends in your .cabal file.
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
   |
11 | import Network.HTTP.Types.Status (statusCode)
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

src/IC/Ref/IO.hs:12:1: error:
    Could not load module ‘Data.CaseInsensitive’
    It is a member of the hidden package ‘case-insensitive-1.2.1.0’.
    Perhaps you need to add ‘case-insensitive’ to the build-depends in your .cabal file.
    Use -v (or `:set -v` in ghci) to see a list of the files searched for.
   |
12 | import Data.CaseInsensitive (original)
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

marcin-dziadus:
> Here are more errors (for `unit-test`):
> 
> ```
> [30 of 49] Compiling IC.Ref.IO        ( src/IC/Ref/IO.hs, dist/build/unit-test/unit-test-tmp/IC/Ref/IO.o, dist/build/unit-test/unit-test-tmp/IC/Ref/IO.dyn_o
> 
> src/IC/Ref/IO.hs:10:1: error:
>     Could not load module ‘Network.HTTP.Client’
>     It is a member of the hidden package ‘http-client-0.6.4.1’.
>     Perhaps you need to add ‘http-client’ to the build-depends in your .cabal file.
>     Use -v (or `:set -v` in ghci) to see a list of the files searched for.
>    |
> 10 | import qualified Network.HTTP.Client as C
>    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 
> src/IC/Ref/IO.hs:11:1: error:
>     Could not load module ‘Network.HTTP.Types.Status’
>     It is a member of the hidden package ‘http-types-0.12.3’.
>     Perhaps you need to add ‘http-types’ to the build-depends in your .cabal file.
>     Use -v (or `:set -v` in ghci) to see a list of the files searched for.
>    |
> 11 | import Network.HTTP.Types.Status (statusCode)
>    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 
> src/IC/Ref/IO.hs:12:1: error:
>     Could not load module ‘Data.CaseInsensitive’
>     It is a member of the hidden package ‘case-insensitive-1.2.1.0’.
>     Perhaps you need to add ‘case-insensitive’ to the build-depends in your .cabal file.
>     Use -v (or `:set -v` in ghci) to see a list of the files searched for.
>    |
> 12 | import Data.CaseInsensitive (original)
>    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ```
> 
> Hmmm, I prepared this:
> 
> ```diff
> $ git diff
> diff --git a/ic-hs.cabal b/ic-hs.cabal
> index 987f767..968c097 100644
> --- a/ic-hs.cabal
> +++ b/ic-hs.cabal
> @@ -400,6 +400,7 @@ test-suite unit-test
>    build-depends: bindings-DSL
>    build-depends: bytestring
>    build-depends: candid
> +  build-depends: case-insensitive
>    build-depends: cborg
>    build-depends: cereal
>    build-depends: containers
> @@ -437,6 +438,7 @@ test-suite unit-test
>    build-depends: winter
>    build-depends: uglymemo
>    build-depends: zlib
> +  other-modules: Data.CaseInsensitive
>    other-modules: IC.Canister
>    other-modules: IC.Canister.Imp
>    other-modules: IC.Canister.Snapshot
> @@ -485,6 +487,8 @@ test-suite unit-test
>    other-modules: IC.Wasm.Imports
>    other-modules: IC.Wasm.Winter
>    other-modules: IC.Wasm.Winter.Persist
> +  other-modules: Network.HTTP.Client
> +  other-modules: Network.HTTP.Types.Status
>  
>  library
>    import: cbits, ghc-flags
> ```
> 
> but it is not enough. I get:
> 
> ```
> [44 of 44] Compiling Main             ( src/ic-ref-run.hs, dist/build/ic-ref-run/ic-ref-run-tmp/Main.o, dist/build/ic-ref-run/ic-ref-run-tmp/Main.dyn_o )
> ld: warning: /nix/store/jidccqzwwm1mgynd434ndxk2gykv9dqz-libiconv-50/lib/libiconv.dylib, ignoring unexpected dylib file
> ld: warning: /nix/store/jidccqzwwm1mgynd434ndxk2gykv9dqz-libiconv-50/lib/libiconv.dylib, ignoring unexpected dylib file
> Linking dist/build/ic-ref-run/ic-ref-run ...
> Preprocessing test suite 'unit-test' for ic-hs-0.0.1..
> Setup: can't find source for Data/CaseInsensitive in src,
> dist/build/unit-test/autogen, dist/build/global-autogen
> ```

Thanks! Excuse my ignorance, but why do those new dependencies need to be listed as `other-modules` and not only `build-depends`?

ggreif:
I am probably just as ignorant, go forward! :-)

marcin-dziadus:
> Are you still blocked here?

No, I think the PR is ready for review.

nomeata (approved):
Pretty neat! Of course, one could write many more tests, but that one always can.

marcin-dziadus:
> Pretty neat! Of course, one could write many more tests, but that one always can.

Agreed, but the specification may still slightly change. Once the final version gets confirmed, I'll add more tests.
@nomeata, thank you for the review!

PR #73, opened 2022-02-18
Bump http-client

marcin-dziadus:
Allow making requests using a raw IPv6 address.


marcin-dziadus:
> LGTM, but just to double-check, won't the updated `all-cabal-hashes` give you the correct version of `http-client` out of the box? Maybe you don't need the version pinning at all...

Thank you, @ggreif! You may be right, let me check...

nomeata:
No,`all-cabal-hashes` does not influence the version picked by nixpkgs; it just allows you to pick different versions. @marcin-dziadus’s first version was exactly how it should be done.

PR #74, opened 2022-03-08
feat: support gzip-compressed canister modules

roman-kashitsyn:
This change adds support for installing gzip-compressed modules
as specified in https://github.com/dfinity/interface-spec/pull/11


marcin-dziadus (approved):
Thanks @roman-kashitsyn!

PR #75, opened 2022-05-02
feat: allow optional "error_code" field in responses

roman-kashitsyn:
This change relaxes the ic-ref-test suite to allow optional "error_code"
field in query responses and in request statuses as described in
https://github.com/dfinity/interface-spec/pull/38

### DIFF
--- a/src/IC/Test/Agent.hs
+++ b/src/IC/Test/Agent.hs
@@ -403,7 +403,7 @@ validateStateCert' what cert = do
 validateStateCert :: (HasCallStack, HasAgentConfig) => Certificate -> IO ()
 validateStateCert = validateStateCert' "certificate"
 
-data ReqResponse = Reply Blob | Reject Natural T.Text
+data ReqResponse = Reply Blob | Reject Natural T.Text (Maybe T.Text)
   deriving (Eq, Show)
 data ReqStatus = Processing | Pending | Responded ReqResponse | UnknownStatus
   deriving (Eq, Show)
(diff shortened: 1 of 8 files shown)


src/IC/Test/Agent.hs
```diff
@@ -416,6 +416,14 @@ prettyBlob x =
   let s = map (chr . fromIntegral) (BS.unpack x) in
   if all isPrint s then s else asHex x
 
+maybeCertValue :: HasCallStack => CertVal a => Certificate -> [Blob] -> IO (Maybe a)
+maybeCertValue cert path = case lookupPath (cert_tree cert) path of
+    Found b -> case fromCertVal b of
+      Just x -> return (Just x)
+      Nothing -> assertFailure $ "Cannot parse " ++ prettyPath path ++ " from " ++ show b
+    Absent -> return Nothing
+    x -> assertFailure $ "Expected to find " ++ prettyPath path ++ ", but got " ++ show x
```
nomeata, line 425: Does the replica support proving absence already? I guess it does in these cases where a whole subtree is requested.

roman-kashitsyn: That was my reasoning as well. It's also a bit embarrassing: my last tree implementation supports proofs of absence; all that's left is to propagate it to the HTTP layer. Sounds like a good weekend project.

roman-kashitsyn:
> It's a bit odd that we don't produce these anywhere in ic-ref, which would be helpful for the agents testing against ic-ref.

Great point! After https://github.com/dfinity/ic-hs/pull/75/commits/8fa86bc3ba73d949ae172fc011417a876cccbe47, ic-ref also emits error codes.

PR #76, opened 2022-05-19
upgrade nixpkgs to release-22.05

basvandijk:
This gives us:

* GHC-9.0.2

* Building now uses less than half the memory as before: from > 12 GB on GHC-8.10.7 to < 6 GB on GHC-9.0.2.

* All dependencies upgraded to their latest versions.

* A working haskell-language-server which allows IDEs like VS Code to analyse ic-hs. For VS Code use the Nix Environment Selector plugin to load `shell.nix` and use the Haskell extension to get:
Screenshot 2022-05-20 at 00 35 56


nix/default.nix
```diff
@@ -42,11 +42,6 @@ let
                 "--set=build.docs=false"
               ];
           });
-
-          all-cabal-hashes = self.fetchurl {
-            url = "https://github.com/commercialhaskell/all-cabal-hashes/archive/d4d63b04cd9f6ed263db8df4cdd6dcc667f96ccd.tar.gz";
-            sha256 = "0pd9zfy8wwbwqg0qkgi28kfi2q3kwl1lwkq2k7b50wm5pb3nngrm";
-          };
```
nomeata, line 49: I’d keep this in, it really helps when we need to bump individual haskell packages beyond what’s in nixpkgs right now. Or at least leave it in as comments.

nix/sources.json
```diff
@@ -32,23 +32,23 @@
...
+        "rev": "94beb7a3edfeb3bcda65fa3f2ebc48ec6b40bf72",
+        "sha256": "065kn8mmpg47dgvg6l4klgy3z9mf4gbrwh25hmswzxc7i20ws70p",
         "type": "tarball",
-        "url": "https://github.com/nmattia/naersk/archive/08afb3d1dbfe016108b72e05b02ba0f6ecb3c8e1.tar.gz",
+        "url": "https://github.com/nmattia/naersk/archive/94beb7a3edfeb3bcda65fa3f2ebc48ec6b40bf72.tar.gz",
         "url_template": "https://github.com/<owner>/<repo>/archive/<rev>.tar.gz"
     },
     "nixpkgs": {
-        "branch": "release-21.11",
+        "branch": "master",
```
nomeata, line 42: It’s nice to stay on a release branch, for a little less churn, and (at least as long as the cachix cache runs in my name) to keep this synced with the `motoko` repo.

Shall we wait until 22.05 is out?

nomeata (commented):
Thanks for helping out here!

Maybe we can merge the compat changes in a separate PR? Or maybe it doesn’t matter.

nix/sources.json
basvandijk: I agree. I started this PR as a preparation for supporting the new nixpkgs release but noticed it was easier than expected so already submitted a draft.

release-22.05 branch off should happen this Sunday (see [release schedule](https://github.com/NixOS/nixpkgs/issues/165792)). I'll adapt my PR to that branch when that has happened.

basvandijk:
> Maybe we can merge the compat changes in a separate PR? 

Yes, I can do that.

nix/sources.json
basvandijk: I'm now tracking `release-22.05`.

nomeata:
Neat! Do you know if Darwin still works?

stefan-at-dfinity (approved):

PR #77, opened 2022-05-19
fix: mergify config

basvandijk:
Mergify changed their `commit_message` option to `commit_message_template`. 
See: https://docs.mergify.com/actions/queue/#id2

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -21,7 +21,7 @@ pull_request_rules:
       queue:
         method: squash
         name: default
-        commit_message_template:
+        commit_message_template: |
           {{title}}
 
           {{body}}
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -21,7 +21,10 @@ pull_request_rules:
       queue:
         method: squash
         name: default
-        commit_message: title+body
+        commit_message_template:
+          {{title}}
+
+          {{body}}
       delete_head_branch: {}
   - name: Clean up automerge tags
     conditions:


.mergify.yml
```diff
@@ -21,7 +21,10 @@ pull_request_rules:
       queue:
         method: squash
         name: default
-        commit_message: title+body
+        commit_message_template:
+          {{title}}
+
+          {{body}}
```
nikolay-komarevskiy, line 27: Did You dump this yml into validator? E.g. https://codebeautify.org/yaml-validator
It shows indent error on line 25. But maybe once the template is expanded there is no more error.

.mergify.yml
```diff
@@ -21,7 +21,10 @@ pull_request_rules:
       queue:
         method: squash
         name: default
-        commit_message: title+body
+        commit_message_template:
```
nomeata, line 24:
```suggestion
        commit_message_template: |
```

nomeata:
Did someone take away my admin rights on this repo :-(?

I think this needs to be force-merged (the mergify status complains about the previous file with `commit_message`), but I can’t do that.

.mergify.yml
nomeata: I think I fixed that.

basvandijk:
> Did someone take away my admin rights on this repo :-(?

If you had "Admin" permissions before then yes because I see you now have "Write" permissions.

.mergify.yml
basvandijk: Thanks for spotting and fixing this folks!

PR #79, opened 2022-05-20
System API for ECDSA signing

marcin-dziadus:
This PR Implements a corresponding part of the IC spec: https://github.com/dfinity/interface-spec/pull/6

It also splits up `IC.Test.Agent` and `IC.Test.Spec` to reduce the excessive memory consumption of compiling those modules.


basvandijk (approved):
LGTM!

I'll help with fixing CI.

marcin-dziadus:
> LGTM!
> 
> I'll help with fixing CI.

Thank you, @basvandijk!

.github/workflows/docs.yml
```diff
@@ -18,6 +18,8 @@ jobs:
         ghc-version: '8.10'
     - name: Build
       run: |
+        sudo apt-get update
+        sudo apt-get install -y libsecp256k1-0 libsecp256k1-dev
```
nomeata, line 22: So far we are using pure Haskell implementations of crypto stuff, for ease of deployment and to distangle reference and production implementation. The `cryptonite` library we are using supports the secp259k1 parameters, <https://hackage.haskell.org/package/cryptonite-0.30/docs/Crypto-PubKey-ECC-Types.html#t:CurveName>. Is there a good reason not to just use that?

src/IC/Ref.hs
```diff
@@ -57,11 +57,18 @@ import qualified Data.Map as M
...
 import qualified Data.ByteString.Lazy as BS
+import qualified Data.ByteString.Short as BSS
 import qualified Data.Text as T
 import qualified Data.Vector as Vec
 import qualified Data.Set as S
+import qualified Data.Binary.Get as Get
+import qualified Data.Binary as Get
+import qualified Haskoin.Keys.Extended as HK
+import qualified Haskoin.Keys.Common as HK
+import qualified Haskoin.Crypto.Signature as HS
```
nomeata, line 68: Non-reference material in `IC.Ref` is something we like to avoid.

I’d suggest to put the plumbing (conversions etc.) into `IC.Crypto.Secp256k1` or `IC.Crypto.Bitcoin` and export functions that will make `IC.Ref` read nicely.

.github/workflows/docs.yml
marcin-dziadus: If I am not mistaken, `cryptonite` does not support key derivation.
We can implement that logic ourselves of course, if you think it would be beneficial.

nomeata: Ah, that may be a reason not use it. Using Haskoin is reasonable then.

nomeata:
Yuck, only now I find that the spec says 

> For curve `secp256k1`, the public key is derived using a generalization of BIP32 (see https://ia.cr/2021/1330[ia.cr/2021/1330, Appendix D]). To derive (non-hardened) https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki[BIP-0032]-compatible public keys, each byte string (`blob`) in the `derivation_path` must be a 4-byte big-endian encoding of an unsigned integer less than 2^31^.

Why is this a `vec { blob }` then, and not a `vec {nat32}`?

In any case, I guess the code here should check this requirement and reject appropriately? And the test suite ought to check that this really fails gracefully?

.github/workflows/docs.yml
basvandijk: We should also switch the docs build to nix but we can do that in another PR.

nomeata: It was neat that, so far, it was possible to develop and build ic-hs without nix. Hence also the tooling that keeps cabal.freeze  (which we don't need for nix). But we can maybe drop that and have more people use nix :-)

The agents and/or dfx used to use that way (without nix, just cabal) of building ic-ref in their CI. We should check if that is still the case.

src/IC/Ref.hs
```diff
@@ -272,6 +275,12 @@ isCanisterRunning cid = getRunStatus cid >>= \case
 isCanisterEmpty :: ICM m => CanisterId -> m Bool
 isCanisterEmpty cid = isNothing . content <$> getCanister cid
 
+getCanisterRootKey :: (CanReject m, ICM m) => CanisterId -> m Bitcoin.ExtendedSecretKey
+getCanisterRootKey cid =
+  gets (M.lookup cid . canisterRootKeys) >>= \case
+    Nothing -> reject RC_CANISTER_ERROR "canister root key does not exist"
```
nomeata, line 281: Can this happen? If it is an invariant that every canister has a key, then this should be an “internal failure”.

But why store that key at all? It seems you can always calculate it using `createExtendedKey`? So this can just be a function?

src/IC/Ref.hs
```diff
@@ -1113,6 +1124,30 @@ runRandIC a = state $ \ic ->
     let (x, g) = runRand a (rng ic)
     in (x, ic { rng = g })
 
+icEcdsaPublicKey :: (ICM m, CanReject m) => EntityId -> ICManagement m .! "ecdsa_public_key"
+icEcdsaPublicKey caller r = do
+    let cid = case (r .! #canister_id) of
+                Just cid -> principalToEntityId cid
+                Nothing -> caller
+    key <- getCanisterRootKey cid
+    case (Bitcoin.derivePublicKey key (Vec.toList (r .! #derivation_path))) of
```
nomeata, line 1133: Minor, but if the goal is to keep this module pretty and `Bitcoin` ugly, maybe let `derivePublicKey` accept a vector, so you don't need to deal with these boring conversions here.

nomeata:
```suggestion
    case Bitcoin.derivePublicKey key (Vec.toList (r .! #derivation_path)) of
```

src/IC/Ref.hs
```diff
@@ -1113,6 +1124,30 @@ runRandIC a = state $ \ic ->
...
+icSignWithEcdsa caller r = do
+    key <- getCanisterRootKey caller
+    case (Bitcoin.derivePrivateKey key (Vec.toList (r .! #derivation_path))) of
+      Left err -> reject RC_CANISTER_ERROR err
+      Right k -> do
+        case (Bitcoin.toHash256 (r .! #message_hash)) of
+          Left err -> reject RC_CANISTER_ERROR err
+          Right h ->
+            return $ R.empty
+              .+ #signature .== (Bitcoin.sign k h)
```
nomeata, line 1149: Just a bunch of unneeded parens:
```suggestion
icSignWithEcdsa :: (ICM m, CanReject m) => EntityId -> ICManagement m .! "sign_with_ecdsa"
icSignWithEcdsa caller r = do
    key <- getCanisterRootKey caller
    case Bitcoin.derivePrivateKey key (Vec.toList (r .! #derivation_path)) of
      Left err -> reject RC_CANISTER_ERROR err
      Right k -> do
        case Bitcoin.toHash256 (r .! #message_hash) of
          Left err -> reject RC_CANISTER_ERROR err
          Right h ->
            return $ R.empty
              .+ #signature .== Bitcoin.sign k h
```

basvandijk:
The `ic-ref-dist` job is failing with:
```
       > Linking dist/build/ic-ref/ic-ref ...
       > /nix/store/l8j77jfvndani23bkp7kax82lsl4lh0c-binutils-2.35.2/bin/ld: cannot find -lsecp256k1
```
I think this is because `secp256k1` doesn't provide a static library (`.a`):
```
$ ls -la $(nix-build nix -A pkgsMusl.secp256k1 --no-link)/lib
total 184
dr-xr-xr-x 3 root root   4096 Jan  1  1970 .
dr-xr-xr-x 4 root root   4096 Jan  1  1970 ..
-r-xr-xr-x 1 root root   1000 Jan  1  1970 libsecp256k1.la
lrwxrwxrwx 1 root root     21 Jan  1  1970 libsecp256k1.so -> libsecp256k1.so.0.0.0
lrwxrwxrwx 1 root root     21 Jan  1  1970 libsecp256k1.so.0 -> libsecp256k1.so.0.0.0
-r-xr-xr-x 1 root root 170872 Jan  1  1970 libsecp256k1.so.0.0.0
dr-xr-xr-x 2 root root   4096 Jan  1  1970 pkgconfig
```
We'll have to see if we can get https://github.com/bitcoin-core/secp256k1 to build statically.

nix/secp256k1/default.nix
```diff
@@ -0,0 +1,47 @@
+{ lib
+, stdenv
+, fetchFromGitHub
+, autoreconfHook
+}:
+
+stdenv.mkDerivation {
+  pname = "secp256k1";
+
+  version = "unstable-2022-05-19";
```
nomeata, line 10: Maybe leave a comment here why we are not using nixpkg's package (version? Flags?), and when we expect to be able to use it again?

basvandijk: Good point. I added a comment in `nix/default.nix` where we override secp256k1.

basvandijk:
Status report
==========

For building `ic-ref-dist` I switched to `pkgsStatic` to get a static `secp256k1`. However `pkgsStatic` removes all dynamic libraries (`.so`). This is problematic for Haskell modules using TemplateHaskell because it requires dynamic libraries to be available.

See: https://github.com/NixOS/nixpkgs/issues/61575

For this reason I had to remove TemplateHaskell support from QuickCheck. But haskell-candid also uses TemplateHaskell which can't be disabled. So that package is failing now.

Not yet sure how to work around this...

basvandijk:
The TemplateHaskell issues have been fixed by patching GHC and disabling haddock in a few libraries.

Now hopefully the last remaining issue is the following error when building `ic-ref-dist`:
```
src/IC/Crypto/BLS.hsc:23 directive strict_import cannot be handled in cross-compilation mode
```

nomeata:
So maybe https://github.com/dfinity/ic-hs/commit/47e1274ea1e11708af792dcbed73488e541801fb cab be cherry-picked?

basvandijk:
I think `IC.Crypto.BLS` is simple enough to replace bindings-DSL with raw hsc.

basvandijk:
> So maybe https://github.com/dfinity/ic-hs/commit/47e1274ea1e11708af792dcbed73488e541801fb cab be cherry-picked?

Done. @nomeata thanks for the pointer!

basvandijk:
Next problem is that `ic-ref-dist` refers to the `warp` Haskell library which is not allowed ([`allowedReferences`](https://nixos.org/manual/nix/stable/expressions/advanced-attributes.html?highlight=allowedReferences#advanced-attributes)`= []`).

The reason is that the `ic-ref` executable from `ic-hs-static` references `warp`:

```
$ grep -a /nix/store/rg01ag16m4kk1xrp6ypw6ypj0wqy5bmk-warp-static-x86_64-unknown-linux-musl-3.3.17 \
          /nix/store/27samy25mh98r4glq2krkmv68l9mxb5a-ic-hs-static-x86_64-unknown-linux-musl-0.0.1/bin/ic-ref
...warp_bindir/nix/store/rg01ag16m4kk1xrp6ypw6ypj0wqy5bmk-warp-static-x86_64-unknown-linux-musl-3.3.17/binwarp_libdir/nix/store/rg01ag16m4kk1xrp6ypw6ypj0wqy5bmk-warp-static-x86_64-unknown-linux-musl-3.3.17/lib/ghc-8.10.7/x86_64-linux-ghc-8.10.7/warp-3.3.17-LFuiV3JNZfpKQMWWUSmbjdwarp_dynlibdir/nix/store/rg01ag16m4kk1xrp6ypw6ypj0wqy5bmk-warp-static-x86_64-unknown-linux-musl-3.3.17/lib/ghc-8.10.7/x86_64-linux-ghc-8.10.7warp_datadir/nix/store/rg01ag16m4kk1xrp6ypw6ypj0wqy5bmk-warp-static-x86_64-unknown-linux-musl-3.3.17/share/x86_64-linux-ghc-8.10.7/warp-3.3.17warp_libexecdir/nix/store/rg01ag16m4kk1xrp6ypw6ypj0wqy5bmk-warp-static-x86_64-unknown-linux-musl-3.3.17/libexec/x86_64-linux-ghc-8.10.7/warp-3.3.17warp_sysconfdir/nix/store/rg01ag16m4kk1xrp6ypw6ypj0wqy5bmk-warp-static-x86_64-unknown-linux-musl-3.3.17/etc...
```

basvandijk:
@nomeata any idea how these paths to warp end up in `ic-ref`?

basvandijk:
Just linking https://github.com/NixOS/nixpkgs/issues/43795#issuecomment-1092029035 here because I think we can simplify our static linking setup when we upgrade to release-22.05 (like removing our GHC override).

nomeata:
> Next problem is that ic-ref-dist refers to the warp Haskell library which is not allowed ([allowedReferences](https://nixos.org/manual/nix/stable/expressions/advanced-attributes.html?highlight=allowedReferences#advanced-attributes)= []).

If these paths are actually unused and don’t affect whether it runs, you can use `removeReferencesTo` to be unblocked here, right?

I guess they are there because warp uses the `Paths_warp` module generated by Cabal, but only use the version number, so probably fine to remove:
```
/tmp/warp-3.3.19 $ grep -r -w Paths_warp 
warp.cabal:                     Paths_warp
warp.cabal:                     Paths_warp
Network/Wai/Handler/Warp/Response.hs:import qualified Paths_warp
Network/Wai/Handler/Warp/Response.hs:warpVersion = showVersion Paths_warp.version
Network/Wai/Handler/Warp/Settings.hs:import qualified Paths_warp
Network/Wai/Handler/Warp/Settings.hs:    , settingsServerName = C8.pack $ "Warp/" ++ showVersion Paths_warp.version
```

basvandijk:
> I guess they are there because warp uses the Paths_warp

On our way to the cinema to see Top Gun my brother @roelvandijk and I came to the same conclusion when discussing this problem ;)

I'll remove the references with `removeReferencesTo`.

PR #81, opened 2022-05-24
Use nixbuild.net as a remote nix builder on CI

basvandijk:
GitHub Actions Runners only have 7GB of memory which is proving to be too small to built ic-hs. So instead of building nix derivations on a runner locally we configure the runner to build derivations remotely on [nixbuild.net](https://nixbuild.net/).

The above requires access to a private SSH key, which is used to authenticate with nixbuild.net. This key is not accessible on PRs originating from forks because that would allow leaking the key. For those PRs we don't configure nixbuild.net as a remote builder and build locally instead.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,45 +11,24 @@ jobs:
       matrix:
         os:
         - ubuntu-latest
-        # TODO: - macos-latest
-        # TODO: - windows-latest
       fail-fast: false
     runs-on: ${{ matrix.os }}
     steps:
     - uses: actions/checkout@v2
-    - name: Configure SSH such that GitHub can authenticate with nixbuild.net
-      run: |
-        sudo mkdir -p -m 700 /root/.ssh
-
-        sudo touch /root/.ssh/nixbuild_id
-        sudo chmod 600 /root/.ssh/nixbuild_id
-        cat << EOF | sudo tee /root/.ssh/nixbuild_id > /dev/null
-        ${{ secrets.SSH_KEY_FOR_NIXBUILD }}
-        EOF
-
-        sudo touch /root/.ssh/config
-        sudo chmod 600 /root/.ssh/config
-        cat << EOF | sudo tee --append /root/.ssh/config > /dev/null
-        Host eu.nixbuild.net
-          PubkeyAcceptedKeyTypes ssh-ed25519
-          IdentityFile /root/.ssh/nixbuild_id
-        EOF
-
-        echo 'eu.nixbuild.net ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIPIQCZc54poJ8vqawd8TraNryQeJnvH1eLpIDgbiqymM' \
-          | sudo tee --append /etc/ssh/ssh_known_hosts > /dev/null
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -11,23 +11,44 @@ jobs:
...
       with:
         extra_nix_config: |
           experimental-features = nix-command
-    - run: nix-env -iA nix-build-uncached -f nix/
     - uses: cachix/cachix-action@v10
       with:
         name: ic-hs-test
         authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+
+    # TODO: we might not need this if we go for remote store builds in nixbuild.net
```
nomeata, line 49: We still need to seed the cache, for the benefit of developers, and I don't think nix-build guarantees that it will cache things for long. Ideally the upload happens from nixbuild directly, but I don't know if that is possible.

rickynils: @nomeata I agree that ideally, nixbuild.net itself should be able to populate the cache, and we are working on implementing that. In the meantime, if you use remote store builds, you could [manually fetch](https://docs.nixbuild.net/remote-builds/#retrieving-build-output-from-remote-stores) the build results from nixbuild.net to the GHA runner, and then either push them explicitly to Cachix, or let the cachix store watcher take care of it.

basvandijk: Looking forward to having nixbuild.net upload directly to the cache!

Regarding using [remote store builds](https://blog.nixbuild.net/posts/2022-03-16-lightning-fast-ci-with-nixbuild-net.html) (which I really would like to switch to) and manually fetching the build results to the GHA runner to upload them to the cache, I don't think we just want to cache the final build results. We actually want to cache all build-time dependencies of the derivation as well. 

But if we fetch the full build-time closure we loose the benefit of remote store builds. So I think it's best to keep using nix-build (or nix-build-uncached) and wait for nixbuild.net to get support for uploading to a cache directly after which we can switch to remote store builds.

basvandijk:
nixbuild.net has built ic-hs and spend less than 5m of CPU time:

```
nixbuild.net> usage

Free build time left: 24:55:45

Resources used since 2022-05-01
-------------------------------
Build count       : 13
Successful builds : 13
Wall time         : 02:06
CPU time          : 04:15
```

yvonneanne (approved):

nomeata:
I guess we can't really test how badly external PR break before this is in master? Maybe we can at least make it fail gracefully and just use local builds then (or maybe the nixbuild GitHub actions are already set up this way)

marcin-dziadus (approved):

basvandijk:
> ... or maybe the nixbuild GitHub actions are already set up this way

@rickynils mentioned in our email thread:

> I should also say that if the nixbuild secret is unavailable to the [nixbuild.net](http://nixbuild.net/) action, it will fail in a rather unhelpful way. We should probably add some option that would simply not configure the remote builder if the secret is missing, falling back to building locally (as Joachim mentions).

It would be great if nixbuild.net action handles missing keys gracefully.

For now I've made the nixbuild.net action conditional on the SSH key being present.

PR #82, opened 2022-05-25
.github/workflows/release.yml: disable MacOS builds in release job

basvandijk:


nomeata (approved):

ericswanson-dfinity:
@basvandijk We use these binaries in the sdk, here: https://github.com/dfinity/sdk/blob/master/nix/sources.json#L95 .   Is there a plan to re-enable this build in the release job?

PR #85, opened 2022-05-30
Mirror changes of the spec related to the canister HTTP calls

marcin-dziadus:


src/IC/Test/Agent/Calls.hs
```diff
@@ -320,6 +326,11 @@ ic_sign_with_ecdsa'' user msg =
 testPort :: HasAgentConfig => Int
 testPort = tc_test_port agentConfig
 
+toTransformFn :: (AllUniqueLabels r1, (r1 .! "function") ~ Candid.FuncRef r2) => 
+                 Maybe String -> Blob -> Maybe (Var r1)
+toTransformFn Nothing _ = Nothing
+toTransformFn (Just name) cid = Just (V.IsJust #function (Candid.FuncRef (Principal cid) (T.pack name)))
```
basvandijk, line 332: This can be simplified using an `fmap` right?

basvandijk (approved):

src/IC/Test/Agent/Calls.hs
marcin-dziadus: Done, thanks!

PR #86, opened 2022-06-16
Revert ".github/workflows/release.yml: disable MacOS builds in release job"

ericswanson-dfinity:
We need to update the ic-ref that we bundle with the SDK to one that supports `ic0.performance_counter`

Reverts dfinity/ic-hs#82


nomeata:
@basvandijk, wdyt?

basvandijk (approved):

PR #87, opened 2022-06-17
fix: macos release

ericswanson-dfinity:
Thanks to a recommendation from @basvandijk 

Previous failure on master: https://github.com/dfinity/ic-hs/runs/6940919295?check_suite_focus=true

Sample successful run on this branch (with the aws s3 cp disabled): https://github.com/dfinity/ic-hs/runs/6942034911?check_suite_focus=true

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -60,12 +60,21 @@ jobs:
           - os: macos-latest
             nix-system: x86_64-darwin
     runs-on: ${{ matrix.os }}
+    env:
+      SSH_KEY_FOR_NIXBUILD: ${{ secrets.SSH_KEY_FOR_NIXBUILD }}
     steps:
     - uses: actions/checkout@v2
-    - uses: nixbuild/nix-quick-install-action@v13
-    - uses: nixbuild/nixbuild-action@v10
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -60,12 +60,21 @@ jobs:
...
+      uses: cachix/install-nix-action@v16
+    - if: matrix.os == 'ubuntu-latest'
+      uses: nixbuild/nix-quick-install-action@v13
+      with:
+        nix_conf: experimental-features = nix-command
+    - if: matrix.os == 'ubuntu-latest' && env.SSH_KEY_FOR_NIXBUILD != ''
+      uses: nixbuild/nixbuild-action@v10
       with:
         nixbuild_ssh_key: ${{ secrets.SSH_KEY_FOR_NIXBUILD }}
+    - run: nix-env -iA nix-build-uncached -f nix/
```
basvandijk, line 77: I realize we don't use `nix-build-uncached` below so we can remove:
```suggestion
```

basvandijk (approved):
Apart from my suggestion it LGTM! Thanks for fixing this!

PR #88, opened 2022-07-13
introduce awaitKnown

stefan-at-dfinity:


src/IC/Test/Agent.hs
```diff
@@ -540,18 +541,23 @@ getRequestStatus sender cid rid = do
       Unknown -> return UnknownStatus
       x -> assertFailure $ "Unexpected request status, got " ++ show x
 
+loop :: HasCallStack => IO (Maybe a) -> IO a
+loop act = go (0::Int)
+  where
+    go 10000 = assertFailure "Polling timed out"
```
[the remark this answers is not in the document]
nomeata: This code is my fault, and was just moved, so this is not a regression from this PR :-)

src/IC/Test/Agent.hs
```diff
@@ -540,18 +541,23 @@ getRequestStatus sender cid rid = do
...
-    loop act = go (0::Int)
-      where
-        go 10000 = assertFailure "Polling timed out"
-        go n = act >>= \case
-          Just r -> return r
-          Nothing -> go (n+1)
+  Responded x -> return $ Just x
+  _ -> return Nothing
+
+awaitKnown :: HasAgentConfig => IO ReqStatus -> IO ReqStatus
```
nomeata, line 557:
```suggestion
-- Polls until status is not Unknown any more, and returns that status
-- even if Pending or Processing
awaitKnown :: HasAgentConfig => IO ReqStatus -> IO ReqStatus
```

mraszyk:
@bogwar confirmed that we need to poll for the status of an asynchronous request after receiving an HTTP response.

src/IC/Test/Agent.hs
```diff
@@ -169,10 +171,11 @@ data AgentConfig = AgentConfig
     , tc_manager :: Manager
     , tc_endPoint :: String
     , tc_test_port :: Int
+    , tc_timeout :: Integer
     }
 
-makeAgentConfig :: String -> Int -> IO AgentConfig
-makeAgentConfig ep' tp = do
+makeAgentConfig :: String -> Int -> Integer -> IO AgentConfig
```
stefan-at-dfinity, line 177: nit: do we ever need more than an Int here? ;-)

mraszyk: Fixed by the new commit.

mraszyk (approved):
Approve after DSD's review.

PR #90, opened 2022-07-28
Draft: reproduce hotfix at commit 25b4c0e04a6eea3516901fbf07e3fc8b039b96ec

mraszyk:


stefan-at-dfinity:
could you add a short PR description?

mraszyk:
This PR reproduces a cycles accounting bug in the mainnet implementation (hotfixed at commit 25b4c0e04a6eea3516901fbf07e3fc8b039b96ec) using ic-ref-test.

It also contains scripts to reproduce the bug before the hotfix and test the hotfix:
```
$ ./init-hotfix.sh
$ ./run-hotfix.sh 63c5693ba56960389e88e3814866e7c99a939f73 # reproduce the bug (see test_63c5693ba56960389e88e3814866e7c99a939f73.txt)
$ ./run-hotfix.sh 50e34926c12aab19d131b8d43b8921881a2a325a # test the fix (see test_50e34926c12aab19d131b8d43b8921881a2a325a.txt)
```
I included a later replica version at commit 50e34926c12aab19d131b8d43b8921881a2a325a because no replica version is available for download from download.dfinity.systems at commit 25b4c0e04a6eea3516901fbf07e3fc8b039b96ec.

It needs to turn off performance counter imports because they were not supported in the replica versions end of april when the bug was discovered and fixed.

The core test logic
```
          _ <- awaitCall cid1 $ callRequest cid1 $ inter_call cid2 "update" defArgs
            { other_side =
              inter_call cid3 "update" defArgs
                {
                  on_reply = acceptAll
                } >>> reply
            , cycles = def_cycles `div` 2
            }
```
is quite compact and thus could have been hit by random testing.

mraszyk:
We are not going to merge this regression test because it is actually a duplicate of an existing regression test. Nevertheless, it shows that random testing using ic-ref-test is a viable technique worth further investigations.

PR #91, opened 2022-09-01
fix: only ascii characters in strings

mraszyk:
Replace all non-ascii characters by their ASCII counterparts to make the code work in the ic repo's nix-shell.


stefan-at-dfinity (approved):

nomeata (commented):
Too bad, I like Unicode :-)

I wonder if the proper fix would be to set `LANG` in the shell.

PR #92, opened 2022-09-02
Request IDs must be 32 bytes in length.

mraszyk:
According to [Interface Spec](https://github.com/dfinity/interface-spec/pull/87), a request id must have length of 32 bytes. Hence, a read request for the status of a request whose id does not have this length is a malformed read state request that should be rejected with 4xx.

### DIFF
--- a/src/IC/Ref.hs
+++ b/src/IC/Ref.hs
@@ -405,7 +405,8 @@ authReadStateRequest t ecid ev (ReadStateRequest user_id paths) = do
               unless (S.member user_id cs) $
                 throwError "User is not authorized to read this metadata field"
             _ -> return () -- public or absent
-      ("request_status":rid: _) -> if BS.length rid /= 32 then throwError "Request IDs must be 32 bytes in length." else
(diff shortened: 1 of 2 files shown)


src/IC/Ref.hs
```diff
@@ -405,7 +405,7 @@ authReadStateRequest t ecid ev (ReadStateRequest user_id paths) = do
               unless (S.member user_id cs) $
                 throwError "User is not authorized to read this metadata field"
             _ -> return () -- public or absent
-      ("request_status":rid: _) ->
+      ("request_status":rid: _) -> if BS.length rid /= 32 then throwError "Request IDs must be 32 bytes in length." else
         gets (findRequest rid) >>= \case
```
nomeata, line 409: I'd use a guard here, instead of `if`

nomeata, line 409:
```suggestion
      ("request_status":rid: _) | BS.length rid /= 32 -> throwError "Request IDs must be 32 bytes in length."
      ("request_status":rid: _) ->                            
        gets (findRequest rid) >>= \case
```

nomeata (approved):

PR #94, opened 2022-10-02
canister http requests test

mraszyk:
Update canister `http_request` functionality in *ic-ref* and *ic-ref-test*:

- replace a local test server by [httpbin](https://httpbin.org/)
- include command line options for specifying httpbin hostname and polling timeout
- do not always reset reject replies from management canister to have reject code 5
- model and test `http_request` fees
- update `http_request` in `ic.did` (add field `max_response_bytes` and variants `post` and `head` to `method`)
- handle https requests in `IC/Ref/IO` using TLS
- update some reject and error codes in ic-ref to match replica


nomeata:
Thanks! A slightly more verbose PR description would be appreciated (also as it may end up as the commit message if it is squashed upon merge)

mraszyk:
> Thanks! A slightly more verbose PR description would be appreciated (also as it may end up as the commit message if it is squashed upon merge)

Done!

nomeata:
I liked the self-containedness of the previous setup; we'll see if this leads to more spurious test failures or trouble in possibly sandboxed builds. But if it makes other things easier, then 🤷‍♂️

basvandijk:
@mraszyk the reason the test fails is that the `coverage` and `ic-ref-test` derivations (CI jobs) are trying to connect to `httpbin.org`. Note that access to the network is disabled in nix-builds to ensure reproducibility. 

I suggest to run a `httbin` locally in the `coverage` and `ic-ref-test` jobs. You can do that quite easily by including a command like the following in the lines above [here](https://github.com/dfinity/ic-hs/blob/master/default.nix#L156) and [here](https://github.com/dfinity/ic-hs/blob/master/default.nix#L173):

```
      ${httpbin} &
```

where you can define the `httpbin` string as something like:

```nix
let
  httpbin =
    let
      python = nixpkgs.python3.withPackages
        (ps: [ ps.httpbin ps.gunicorn ps.gevent ]);
    in "${python}/bin/gunicorn -b 0.0.0.0:8003 httpbin:app -k gevent";
in
```

stefan-at-dfinity (approved):

PR #95, opened 2022-10-02
drop metadata.absent test

mraszyk:
Interface Spec does not guarantee that the path `/canister/<canister_id>/metadata/<name>` may be requested if the custom section `<name>` does not exist. Unlike for paths with prefix `/request_status/<request_id>`, providing a valid absence proof for paths `/canister/<canister_id>/metadata/<name>` might reveal private custom section names which might not be desirable.

### DIFF
--- a/src/IC/Test/Spec.hs
+++ b/src/IC/Test/Spec.hs
@@ -1643,10 +1643,7 @@ icTests = withAgentConfig $ testGroup "Interface Spec acceptance tests"
             where sized x = BS.fromStrict (toLEB128 @Natural (fromIntegral (BS.length x))) <> x
           withSections xs = foldl withCustomSection trivialWasmModule xs
       in
-      [ simpleTestCase "absent" $ \cid -> do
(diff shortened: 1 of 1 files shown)


nomeata:
I would not guide the design by the desire to hide existing fields in the hash tree. That data structure is very much unsuitable to hide data (e.g. an absence proof of a neighbouring value can easily reveal existence. Maybe not in this case, but in general. Also hashes of parent nodes are known, and certainly leak data). The only secrecy, if any at all, I'd rely on is through a cryptographic (and salted) hash in the leafs.

With that aside, it seems that returning an absence proof is more useful.

It's also consistent with request status handling, where anyone can ask for an absence proof, but only authorized readers can get a value (this is a bit odd, but I don't see a better way).

So I'd suggest to adjust the spec to say that this ought to return an absence proof.

mraszyk:
@nomeata Thank you for sharing your thoughts on this! We decided to follow your advice so I'm closing this PR now.

PR #99, opened 2022-10-21
remove tests enforcing effective canister id checks

mraszyk:
This test is unlikely to ever pass the replica implementation.


nomeata:
If that's the case, it may be worth tightening the spec to the status quo.

mraszyk:
> If that's the case, it may be worth tightening the spec to the status quo.

https://github.com/dfinity/interface-spec/pull/101/commits/50577f94e41994c85c5f756668ed94adbc8c3f12

mraszyk:
resolved

PR #101, opened 2022-11-02
add transform context for canister http_request

mraszyk:
This PR makes the following changes:
- add context to transform for canister http_request and update the canister http_request fee accordingly
- the caller of transform function for canister http_request is the management canister
- add tests for the above two changes

The corresponding Interface Spec can be found [here](https://github.com/dfinity/interface-spec/pull/96).

### DIFF
--- a/universal-canister/src/main.rs
+++ b/universal-canister/src/main.rs
@@ -1,6 +1,48 @@
 use std::convert::TryInto;
 
+use candid::{CandidType, Deserialize, Decode, Encode, Principal};
+use serde::Serialize;
+
 mod api;
+
+// Canister http_request types
+
+#[derive(CandidType, Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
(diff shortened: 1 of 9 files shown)


universal-canister/src/main.rs
```diff
@@ -304,8 +346,15 @@ fn eval(ops: Ops) {
                 api::call_cycles_add128(high, low)
             }
 
-            // canister heartbeat script
+            // canister http_request
```
stefan-at-dfinity, line 349: A version of the UC is hardcoded in the `ic` repository. If we make such changes, let's also follow up with an MR that updates the UC in `ic`. wdyt?

stefan-at-dfinity (approved):

universal-canister/src/main.rs
mraszyk: Sure, we just need to make sure that `ic-cdk-optimizer` is called on it because otherwise it cannot be installed on application subnets.

PR #102, opened 2022-11-02
add subnets to reference implementation

mraszyk:
This PR brings the reference implementation closer to the mainnet implementation by modeling subnets. In more detail, it makes the following changes:
- IC record contains a list of subnets, each characterized by its id, type, key pair, and canister ranges
- `read_state` requests can fail if the effective canister id does not belong to any subnet
- `create_canister` creates a canister on the caller's subnet
- `provisional_create_canister_with_cycles` creates a canister on the subnet of the effective canister id (if called by a user) or on the caller's subnet (otherwise)
- canister creation can fail if the subnet has surpassed its canister id allocation
- the initial subnet configurations are specified when starting ic-ref (keys are derived from the provided nonces and subnet ids are derived from the public keys)
- `mkOpaqueId` appends two `0x01` bytes to a blob (in accordance with the replica implementation)
- opaque canister ids are derived from integers via their big-endian representation so that the order on integers matches the order on the derived canister ids
- `freshId`s are generated from the given canister ranges (of a subnet)
- `rootSubnet` (first system subnet in the initial subnet configurations) does not include delegations in its certificates, other subnets include delegations in their certificates
- user requests are stored with their effective canister id to implement access control for request_status in read_state requests according to the spec
- include `/subnet` subtree in read_state responses (if requested)
- ic-ref-test takes an effective canister id for user requests to selected methods of the management canister as a parameter
- ic-ref-test handles failing read_state requests due to effective canister id checks


src/IC/Ref.hs
```diff
@@ -581,18 +596,24 @@ delegationTree (Timestamp t) (EntityId subnet_id) subnet_pub_key = node
...
     (=:) = M.singleton
 
-getPrunedCertificate :: ICM m => Timestamp -> [Path] -> m Certificate
-getPrunedCertificate time paths = do
+getSubnetFromCanisterId :: (CanReject m, ICM m) => CanisterId -> m (EntityId, SubnetType, SecretKey, [(W.Word64, W.Word64)])
+getSubnetFromCanisterId cid = do
+    subnets <- gets subnets
+    case find (\(_, _, _, ranges) -> find (\(a, b) -> wordToId a <= cid && cid <= wordToId b) ranges /= Nothing) subnets of
+      Nothing -> reject RC_SYS_FATAL "Canister id does not belong to any subnet." Nothing
+      Just x -> return x
```
stefan-at-dfinity, line 604: suggestion to make this a bit more readable.

```suggestion
getSubnetFromCanisterId cid = do
    subnets <- gets subnets
    case subnetOfCid cid subnets of
      Nothing -> reject RC_SYS_FATAL "Canister id does not belong to any subnet." Nothing
      Just x -> return x
    where
      subnetOfCid cid subnets = find (\(_, _, _, ranges) -> find (\(a, b) -> wordToId a <= cid && cid <= wordToId b) ranges /= Nothing) subnets
```

stefan-at-dfinity (approved):

basvandijk (approved):

PR #106, opened 2022-11-17
query balance and balance128 within the same query

mraszyk:
Querying the balance of a canister with installed code twice in a row might not yield the same result because the canister is charged for resource usage:
```
    cycles API - backward compatibility
      canister_cycle_balance = canister_cycle_balance128 for numbers fitting in 64 bits: FAIL (10.40s)
        src/IC/Test/Agent.hs:666:
        expected: 1152921502679711556
         but got: 1152921502680301640
```
This PR fixes the above test by querying balance and balance128 within the same query.

### DIFF
--- a/ic-hs.cabal
+++ b/ic-hs.cabal
@@ -449,7 +449,6 @@ test-suite unit-test
   build-depends: unordered-containers
   build-depends: utf8-string
   build-depends: vector
-  build-depends: wide-word
   build-depends: winter
   build-depends: uglymemo
   build-depends: zlib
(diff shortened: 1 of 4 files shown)


basvandijk (approved):
Maybe we should consider using [Word128](https://hackage.haskell.org/package/wide-word-0.1.1.2/docs/Data-WideWord-Word128.html) for 128bit words instead of `Natural`.

mraszyk:
> Maybe we should consider using [Word128](https://hackage.haskell.org/package/wide-word-0.1.1.2/docs/Data-WideWord-Word128.html) for 128bit words instead of `Natural`.

Good point! Done.

PR #107, opened 2022-11-17
Add canister global timer

mraszyk:
Add canister global timer specified [here](https://github.com/dfinity/interface-spec/pull/111).


src/IC/Canister/Imp.hs
```diff
@@ -585,6 +612,18 @@ systemAPI esref =
...
+    global_timer_set ts = do
+        old_timer <- gets new_global_timer
+        case old_timer of
+          Nothing -> do
+            old_timer <- gets (env_global_timer . env)
+            puts (\es -> es {new_global_timer = Just ts})
+            return $ fromIntegral old_timer
+          Just old_timer -> do
+            puts (\es -> es {new_global_timer = Just ts})
+            return old_timer
```
basvandijk, line 625: Since you `puts` the same state in both branches why not hoist it outside the case like:

```suggestion
    global_timer_set :: Word64 -> HostM s Word64
    global_timer_set ts = do
        old_timer <- gets new_global_timer
        puts $ \es -> es {new_global_timer = Just ts}
        case old_timer of
          Nothing -> fromIntegral <$> gets (env_global_timer . env)
          Just old_timer -> return old_timer
```
```

File: src/IC/Canister/Imp.hs [hunk 2]
```diff
@@ -185,70 +190,92 @@ subtractAvailable esref f = do
...
-  , toImport "ic0" "msg_method_name_copy" msg_method_name_copy
-
-  , toImport "ic0" "accept_message" accept_message
-  , toImport "ic0" "time" get_time
-  , toImport "ic0" "performance_counter" performance_counter
-  , toImport "ic0" "canister_state_counter" get_canister_state_counter
-
-  , toImport "ic0" "debug_print" debug_print
-  , toImport "ic0" "trap" explicit_trap
+  [ toImport' "ic0" "msg_arg_data_size" "I U Q Ry F" msg_arg_data_size
```
basvandijk, line 193: What are these contexts? A comment somewhere could help readers a lot. 

And wouldn't it be safer to represent them as a list instead of a String using something like:
```suggestion
  [ toImport' "ic0" "msg_arg_data_size" [I, U, Q, Ry, F] msg_arg_data_size
```

basvandijk (commented):
Reviewed 1/3 so far. Will review the rest tomorrow.

src/IC/Ref.hs
```diff
@@ -1027,6 +1041,7 @@ icInstallCode caller r = do
             }
         performCanisterActions canister_id ca
         bumpCanisterStateCounter canister_id
+        if set_global_timer ca == Nothing then setCanisterGlobalTimer canister_id 0 else return ()
```
basvandijk, line 1044:
```suggestion
        when (set_global_timer ca == Nothing) $ setCanisterGlobalTimer canister_id 0
```

src/IC/Ref.hs
```diff
@@ -1044,16 +1059,17 @@ icInstallCode caller r = do
...
           `onTrap` (\msg -> reject RC_CANISTER_ERROR ("Post-upgrade trapped: " ++ msg) (Just EC_CANISTER_TRAPPED))
 
         setCanisterContent canister_id $ CanisterContent
             { can_mod = new_can_mod
             , wasm_state = new_wasm_state
             }
-        performCanisterActions canister_id (ca1 <> ca2)
+        performCanisterActions canister_id (ca1 { set_global_timer = Nothing } <> ca2)
         bumpCanisterStateCounter canister_id
+        if set_global_timer ca2 == Nothing then setCanisterGlobalTimer canister_id 0 else return ()
```
basvandijk, line 1072:
```suggestion
        when (set_global_timer ca2 == Nothing) $ setCanisterGlobalTimer canister_id 0
```

src/IC/Ref.hs
```diff
@@ -1339,10 +1361,33 @@ runHeartbeat cid = do
...
-processHeartbeats :: ICM m => m ()
-processHeartbeats = do
+runGlobalTimer :: ICM m => CanisterId -> m ()
+runGlobalTimer cid = do
+  is_empty   <- isCanisterEmpty cid
+  is_running <- isCanisterRunning cid
+  Timestamp current_time <- getCanisterTime cid
+  global_timer <- getCanisterGlobalTimer cid
+  let should_fire = global_timer /= 0 && current_time >= global_timer
+  unless (is_empty || not is_running || not should_fire) $ do
```
basvandijk, line 1371: Personally reversing the logic is more understandable to me since you need fewer double negations:
```suggestion
  when (should_fire && is_running && not is_empty) $ do
```
But that's personal taste so ignore if you don't agree.

src/IC/Test/Spec.hs
```diff
@@ -49,6 +51,18 @@ import IC.Test.Agent.Calls
...
+
+loop :: HasAgentConfig => IO (Maybe a) -> IO a
+loop act = getCurrentTime >>= go
+  where
+    go init = act >>= \case
+      Just x -> return x
+      Nothing -> do
+        now <- getCurrentTime
+        if diffUTCTime now init > fromIntegral (tc_timeout agentConfig) then assertFailure "Polling timed out"
+        else go init
```
basvandijk, line 64: It seems you're not really using the `a` in `Maybe a` in the single callsite of `loop` so why not replace it with a `Bool`?

Additionally, continually checking for the time is less efficient and arguably less understandable than using [timeout](https://hackage.haskell.org/package/base-4.17.0.0/docs/System-Timeout.html#v:timeout).

So why not something like:

```suggestion
waitFor :: HasAgentConfig => IO Bool -> IO ()
waitFor act = do 
    result <- timeout (tc_timeout agentConfig * 10 ^ 6) loop
    when (isNothing result) $ assertFailure "Polling timed out"
  where
    loop = do 
      stop <- act
      unless stop loop      
```

You should also consider adding in a short [`threadDelay`](https://hackage.haskell.org/package/base-4.17.0.0/docs/Control-Concurrent.html#v:threadDelay) before the `loop` in `unless stop loop` to not create to hot a loop.

src/IC/Test/Spec.hs
```diff
@@ -1199,6 +1215,114 @@ icTests = withAgentConfig $ testGroup "Interface Spec acceptance tests"
...
+    let get_global cid = call cid (replyData $ getGlobal) in
+    let get_current_time = floor . (* 1e9) <$> getPOSIXTime in
+    let get_far_past_time = return 1 in
+    let get_far_future_time = floor . (* 1e9) <$> (+) 100000 <$> getPOSIXTime in
+    let set_timer cid time = call cid (replyData $ i64tob $ apiGlobalTimerSet $ int64 time) in
+    let blob = toLazyByteString . word64LE . fromIntegral in
+    let wait_for_timer cid n = loop $ do
+          ctr <- get_global cid
+          return $ if ctr == blob n then Just () else Nothing
+        in
```
basvandijk, line 1230: Using the above `waitFor` this could become:
```suggestion
    let wait_for_timer cid n = waitFor $ (blob n ==) <$> get_global cid in
```

src/IC/Canister/Imp.hs
mraszyk: Great suggestion!

basvandijk (approved):
Looks great!

PR #110, opened 2022-11-21
update release.yml

mraszyk:


.github/workflows/release.yml
```diff
@@ -48,6 +48,16 @@ jobs:
     - run: nix-build-uncached -A check-cabal-freeze
     # now the rest
     - run: nix-build-uncached
+    - id: buildit
+      run: |
+        out_path="$(nix-build -A ic-ref-dist)"
+        echo "::set-output name=out::$out_path"
```
ggreif, line 54: `::set-output` is deprecated, see

mraszyk: I just copied this over from the bottom of the file. Closing because I accidentally opened an PR before testing out.

PR #113, opened 2022-11-24
run system tasks periodically

mraszyk:
Currently, system tasks (heartbeat and global timer) are only triggered upon receiving an HTTP request. This MR adds a new execution thread that periodically triggers these tasks every second.

### DIFF
--- a/src/IC/HTTP.hs
+++ b/src/IC/HTTP.hs
@@ -27,10 +27,14 @@ import IC.Crypto
 
 withApp :: [SubnetConfig] -> Maybe FilePath -> (Application -> IO a) -> IO a
 withApp subnets backingFile action =
-    withStore (initialIC subnets) backingFile $ \store -> forkIO (loopIC store) >> (action $ handle store)
+    withStore (initialIC subnets) backingFile $ \store -> 
+      withAsync (loopIC store) $ \_async -> 
+        action $ handle store
   where
     loopIC :: Store IC -> IO ()
-    loopIC store = modifyStore store aux >> threadDelay 1000000 >> loopIC store
+    loopIC store = forever $ do
+        modifyStore store aux
+        threadDelay 1000000
       where
         aux = do
           lift getTimestamp >>= setAllTimesTo
--- a/src/IC/HTTP.hs
+++ b/src/IC/HTTP.hs
@@ -93,7 +93,11 @@ handle store req respond = case (requestMethod req, pathInfo req) of
   where
     runIC :: StateT IC IO a -> IO a
     runIC a = do
-      x <- modifyStore store a
+      x <- modifyStore store $ do
+        -- Here we make IC.Ref use “real time”
+        lift getTimestamp >>= setAllTimesTo
+        processSystemTasks
+        a
       -- begin processing in the background (it is important that
       -- this thread returns, else warp is blocked somehow)
       void $ forkIO loopIC
--- a/src/IC/HTTP.hs
+++ b/src/IC/HTTP.hs
@@ -30,7 +30,7 @@ withApp subnets backingFile action =
     withStore (initialIC subnets) backingFile $ \store -> forkIO (loopIC store) >> (action $ handle store)
   where
     loopIC :: Store IC -> IO ()
-    loopIC store = modifyStore store aux >> threadDelay 1000 >> loopIC store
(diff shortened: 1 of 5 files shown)


src/IC/HTTP.hs
```diff
@@ -27,7 +27,14 @@ import IC.Crypto
...
 withApp subnets backingFile action =
-    withStore (initialIC subnets) backingFile (action . handle)
+    withStore (initialIC subnets) backingFile $ \store -> forkIO (loopIC store) >> (action $ handle store)
+  where
+    loopIC :: Store IC -> IO ()
+    loopIC store = modifyStore store aux >> threadDelay 1000000 >> loopIC store
+      where
+        aux = do
+          lift getTimestamp >>= setAllTimesTo
+          processSystemTasks
```
basvandijk, line 37: It would be better if the lifetime of the thread is tied to the lifetime of the action. So what about using the [async](https://hackage.haskell.org/package/async) here and have:

```suggestion
    withStore (initialIC subnets) backingFile $ \store -> 
      withAsync (loopIC store) $ \_async -> 
        action $ handle store
  where
    loopIC :: Store IC -> IO ()
    loopIC store = forever $ do
        modifyStore store aux
        threadDelay 1000000
      where
        aux = do
          lift getTimestamp >>= setAllTimesTo
          processSystemTasks
```

Note that I also redefined `loopIC` in terms of [`forever`](https://hackage.haskell.org/package/base-4.17.0.0/docs/Control-Monad.html#v:forever)

nomeata:
I’m worried that using threads here rather than deterministic inputs will make it harder to use `ic-ref` as a deterministic environment to write test, as we do in Motoko. (Ok, there we use `ic-ref-test`, which could do something else). Just a thought.

mraszyk:
> I’m worried that using threads here rather than deterministic inputs will make it harder to use `ic-ref` as a deterministic environment to write test, as we do in Motoko. (Ok, there we use `ic-ref-test`, which could do something else). Just a thought.

My motivation to write this PR was actually that @ggreif could not use `ic-ref` to test the global timers in Motoko.

mraszyk:
@nomeata Would it make sense to you to only start this thread if a corresponding flag is passed?

nomeata:
That might help indeed. And for `ic-ref-test` (and ideally also `drun`) I would suggest a new “skript command” command “run heartbeats now”, maybe even for a specific canister, to give the test harnesses good control over when the heartbeat runs. This allows to reliably check, for example, how a canister behaves if a heatbeat does _or does not_ run between two other events.

mraszyk:
> And for `ic-ref-test` (and ideally also `drun`) I would suggest a new “skript command” command “run heartbeats now”, maybe even for a specific canister, to give the test harnesses good control over when the heartbeat runs. This allows to reliably check, for example, how a canister behaves if a heatbeat does _or does not_ run between two other events.

This suggestion would not work for @ggreif 's test because the global timer would need to run *during* the execution of an update call, i.e., after it's submitted and before it returns.

nomeata:
Ah, yes, the old problem of how to influence the message scheduler from tests. It already ariess without heartbeats, if one wants to test various interleavings.

With @ielashi I discussed a possible approach a while ago, with barriers that hold back responses to messages, and thus allow you to do other things in between two specific messages, and I think he even had a prototype, but it never landed.

I am wary of trying to solve “we need better control over the scheduling to run specific tests” with “let’s introduce concurrency and timer-based interaction and hope for the best”, which I fear will lead to flaky (timing sensitive) and slower (because you need to poll or even have fixed delays) tests.

I don’t have much to offer, though, besides the barrier based approach (maybe @ielashi can share documents we wrote back then, I don’t have access. Might have been an issue on the old `ic-ref` repo.)

mraszyk:
Steering system tasks in `ic-ref` is difficult without exposing a side-channel. In `ic-ref-run`, it's easier, but there the IC is not under heavy load (every test has its own IC instance) so the tests should not be flaky. I think it makes sense to add a flag specifying how often you want to run heartbeats. This should give enough determinism to the tests (although it could make their execution slower).

ggreif:
Heads-up: a tail-call test now times out in the Motoko test suite under `ic-ref`...

ggreif (approved):
Both `heartbeat` as well as `global_timer` work now, I am happy!

PR #114, opened 2022-11-28
ic-ref-run: Execute heartbeats before any submitted message

mraszyk:
Follow-up on #69


ggreif:
Will this place the canister callback invocations nearer to the scheduled expiration date?

mraszyk:
> Will this place the canister callback invocations nearer to the scheduled expiration date?

I didn't quite get the question. The change would make heartbeats and timers more deterministic by triggering them if new calls are submitted.

ggreif:
> > Will this place the canister callback invocations nearer to the scheduled expiration date?
> 
> I didn't quite get the question. The change would make heartbeats and timers more deterministic by triggering them if new calls are submitted.

Yeah, this answers my question.

mraszyk:
@ggreif Could you please make sure this MR does not break your tests and approve if it works?

ggreif:
I don't see much of a difference. The tests give the same results. I think that is a good sign.

ggreif (approved):
Looks like a sensible thing to do.

PR #115, opened 2022-11-29
do not return an HTTP error for calls to stopping/stopped canisters

mraszyk:
Calls to stopping/stopped canisters should not be rejected with an HTTP error (see [API Request submission](https://ic-interface-spec.netlify.app/#_api_request_submission) which can succeed even if the canister is not running). Instead, the request should go through the states received, processing, and finally rejected.


nomeata (approved):

mraszyk:
@nomeata Thank you for taking a look!

PR #121, opened 2022-12-06
allow a canister to query its own status

mraszyk:
This PR implements the following change of Interface Spec: https://github.com/dfinity/interface-spec/pull/91


dsd-at-dfinity (approved):
thanks!

PR #122, opened 2022-12-07
fix macOS artifacts

mraszyk:
This MR fixes the error
```
dyld[2382]: Library not loaded: '@executable_path/libz.1.2.12.dylib'
  Referenced from: '/Users/runner/.cache/dfinity/versions/0.12.1+rev36.7883e5aa/ic-ref'
  Reason: tried: '/Users/runner/.cache/dfinity/versions/0.12.1+rev36.7883e5aa/libz.1.2.12.dylib' (no such file), '/usr/local/lib/libz.1.2.12.dylib' (no such file), '/usr/lib/libz.1.2.12.dylib' (no such file)
```
observed on macOS when using our artifact.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -50,7 +50,7 @@ jobs:
     - run: nix-build-uncached
 
   release:
-    if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' }}
+    if: ${{ github.event_name == 'push' }}
     needs: test
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -91,7 +91,12 @@ jobs:
         echo "::set-output name=out::$out_path"
 
     - run: |
-        tar -C "${{ steps.buildit.outputs.out }}/bin" -czvf ic-ref.tar.gz ic-ref
+        if [[ -d "${{ steps.buildit.outputs.out }}/bin/libs" ]]
```
mraszyk, line 94: I know this is ugly, but didn't find a better way to make `tar` work given that the directory `libs` only exists for macOS builds.

basvandijk (approved):
It feels a bit off to have a `lib` directory in the `bin` directory but if it works it works.

PR #123, opened 2022-12-11
scale cycle cost of http_request calls based on subnet size

mraszyk:
The replica implementation scales the cost of calls as follows:
```
    // Scale cycles cost according to a subnet size.
    fn scale_cost(&self, cycles: Cycles, subnet_size: usize) -> Cycles {
        match self.use_cost_scaling_flag {
            false => cycles,
            true => (cycles * subnet_size) / self.config.reference_subnet_size,
        }
    }
```
where `self.config.reference_subnet_size` is set to
```
/// Default subnet size which is used to scale cycles cost according to a subnet replication factor.
///
/// All initial costs were calculated with the assumption that a subnet had 13 replicas.
/// IMPORTANT: never set this value to zero.
const DEFAULT_REFERENCE_SUBNET_SIZE: usize = 13;
```
Recently, `self.use_cost_scaling_flag` was enabled and thus this PR implements cycle cost scaling based on subnet size also in the reference implementation.

The scaling is only implemented for canister `http_request` management canister calls because no other operations are being charged for in the reference implementation (yet).

### DIFF
--- a/src/IC/Constants.hs
+++ b/src/IC/Constants.hs
@@ -14,6 +14,10 @@ import IC.Types
 cDEFAULT_PROVISIONAL_CYCLES_BALANCE :: Natural
 cDEFAULT_PROVISIONAL_CYCLES_BALANCE = 100_000_000_000_000
 
+-- Subnets
+reference_subnet_size :: W.Word64
+reference_subnet_size = 13
+
 -- Canister http_request limits
 max_request_bytes_limit :: W.Word64
 max_request_bytes_limit = 2_000_000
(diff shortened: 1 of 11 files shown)


src/IC/Constants.hs
```diff
@@ -14,6 +14,10 @@ import IC.Types
 cDEFAULT_PROVISIONAL_CYCLES_BALANCE :: Natural
 cDEFAULT_PROVISIONAL_CYCLES_BALANCE = 100_000_000_000_000
 
+-- Subnets
+reference_subnet_size :: W.Word64
+reference_subnet_size = 13
```
dsd-at-dfinity, line 19: For completeness, one might add that setting this to `0` is invalid/will cause runtime errors.

PR #125, opened 2022-12-12
test setting timer in timer callback

mraszyk:


src/IC/Test/Spec.hs
```diff
@@ -1689,6 +1689,16 @@ icTests sub = withAgentConfig $ testGroup "Interface Spec acceptance tests"
       timer1 @?= blob 0
       timer2 @?= blob 0
       timer3 @?= blob far_far_future_time
+    , testCase "in timer callback" $ do
+      past_time <- get_far_past_time
+      far_future_time <- get_far_future_time
+      cid <- install $ onGlobalTimer $ callback $ set_timer_prog far_future_time -- timer callback sets timer to far_future_time and sets the previous value to global
+      _ <- reset_global cid -- sets global to 42
+      timer1 <- set_timer cid past_time -- sets timer to 1 and returns previous (0) timer
```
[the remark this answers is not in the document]
mraszyk: This line
```
timer1 <- set_timer cid past_time
```
does not execute anything in the callback. It just schedules the callback at a far past time and once it gets executed, the program `callback $ set_timer_prog far_future_time` from
```
cid <- install $ onGlobalTimer $ callback $ set_timer_prog far_future_time
```
is run.

src/IC/Test/Spec.hs
```diff
@@ -1689,6 +1689,16 @@ icTests sub = withAgentConfig $ testGroup "Interface Spec acceptance tests"
       timer1 @?= blob 0
       timer2 @?= blob 0
       timer3 @?= blob far_far_future_time
+    , testCase "in timer callback" $ do
+      past_time <- get_far_past_time
+      far_future_time <- get_far_future_time
+      cid <- install $ onGlobalTimer $ callback $ set_timer_prog far_future_time -- timer callback sets timer to far_future_time and sets the previous value to global
```
ggreif, line 1695: `callback $ set_timer_prog far_future_time` overwrites the global timer in the callback but doesn't check that the previous value of the global timer is 0. Shouldn't the callback also check that?

mraszyk: the previous value of the global timer is written to the global variable in the callback and the value of zero (for the global variable) is awaited on line `wait_for_timer cid 0`

ggreif (approved):

PR #129, opened 2022-12-23
RUN-509: Enable Canister Timers in pre-upgrade

dfinity-berestovskyy:
Canister Timers are allowed in post-upgrade and init and it is easy to assume that they are also okay in pre-upgrade.


mraszyk (commented):
The CI fails and I suspect this is because you opened a MR from a fork. Would you mind pushing your branch to the ic-hs repo and open an MR from there?

mraszyk:
I've opened a PR #134 duplicating this one on a branch that is not on a fork.

PR #136, opened 2023-01-24
fix conditions on read_state requests

mraszyk:
This MR implements the following fixes of `ic-ref`:
- implement access control for `read_state` requests according to [Interface Spec](https://ic-interface-spec.netlify.app/#http-read-state)
- `create_canister` cannot be called via ingress message

### DIFF
--- a/src/IC/Test/Agent.hs
+++ b/src/IC/Test/Agent.hs
@@ -85,6 +85,7 @@ module IC.Test.Agent
       isPendingOrProcessing,
       isReject,
       isReply,
+      isResponded,
       okCBOR,
       otherSK,
       otherUser,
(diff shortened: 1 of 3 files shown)


src/IC/Test/Agent.hs
```diff
@@ -569,6 +570,10 @@ getRequestStatus' sender cid rid = do
 getRequestStatus :: (HasCallStack, HasAgentConfig) => Blob -> Blob -> Blob -> IO ReqStatus
 getRequestStatus sender cid rid = getRequestStatus' sender cid rid >>= is2xx
 
+isResponded :: ReqStatus -> Assertion
+isResponded (Responded _) = return ()
+isResponded _ = assertFailure "Request must be replied"
```
nomeata, line 575: It checks for responded (reply or reject), but the error message says replied. Which one is it?

dsd-at-dfinity (approved):

PR #146, opened 2023-02-27
.github/workflows/release.yml: skip coverage job on darwin

basvandijk:
The GitHub darwin runners only have 14GB of memory. However the `coverage` job requires 17GB which means those machines start swapping which grinds exection to a halt eventually timing out the job.

Since we only care about building the `ic-ref-dist` job on darwin we only run the problematic `coverage` job on linux.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,9 +39,12 @@ jobs:
     # run a few targets explicitly, to get easier signal in the CI view
     - run: nix-build-uncached -A universal-canister
     - run: nix-build-uncached -A ic-hs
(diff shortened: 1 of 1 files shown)


mraszyk:
"ic-ref-test" job/derivation also runs some tests (i.e., expected to trigger 17G again) that are covered by the "coverage" job so let's maybe just drop "ic-ref-test"

mraszyk:
(hopefully) finally, the last CI job nix-build-uncached seems to again build ic-ref-test

mraszyk (approved):
The MR #138 is green, so I'm approving. Thanks a lot!

PR #148, opened 2023-02-28
Upgrade to GHC-9.4 resulting in lower compile-time memory usage

basvandijk:
Building ic-hs with GHC-9.0 requires ~13GB of memory. Upgrading to GHC-9.4 reduces this ~9GB:

Screenshot 2023-02-28 at 17 21 31

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -24,31 +24,42 @@ let universal-canister = (naersk.buildPackage rec {
 }); in
 
 let haskellOverrides = self: super:
-    let generated = import nix/generated/all.nix self super; in
+    let generated = import nix/generated/all.nix self super;
+        inherit (nixpkgs.haskell.lib) doJailbreak;
+    in
     generated //
     {
+      wide-word = super.callHackage "wide-word" "0.1.5.0" {};
       haskoin-core = nixpkgs.haskell.lib.dontCheck (nixpkgs.haskell.lib.markUnbroken super.haskoin-core);
+      microlens-platform = doJailbreak super.microlens-platform;
+      base16 = doJailbreak super.base16;
+      base32 = doJailbreak super.base32;
+      candid = doJailbreak generated.candid;
     }; in
 
-let haskellPackages = nixpkgs.haskellPackages.override {
+let haskellPackages = nixpkgs.haskell.packages.ghc943.override {
   overrides = haskellOverrides;
 }; in
 
-let staticHaskellPackages = nixpkgs.pkgsStatic.haskellPackages.override {
+let staticHaskellPackages = nixpkgs.pkgsStatic.haskell.packages.ghc943.override {
   overrides = haskellOverrides;
 }; in
 
 let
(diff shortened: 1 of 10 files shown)


mraszyk (dismissed):
LGTM

.github/workflows/release.yml
```diff
@@ -18,6 +18,15 @@ jobs:
       SSH_KEY_FOR_NIXBUILD: ${{ secrets.SSH_KEY_FOR_NIXBUILD }}
     steps:
     - uses: actions/checkout@v3
+    - run: |
+        mkdir "$HOME/.ssh"
+        chmod 700 "$HOME/.ssh"
+        cat << EOF > "$HOME/.ssh/config"
+        Host *
+          ServerAliveInterval 300
+          ServerAliveCountMax 2
```
basvandijk, line 27: Let's try out:
```suggestion
          ServerAliveCountMax 0
```

.github/workflows/release.yml
```diff
@@ -18,6 +18,15 @@ jobs:
       SSH_KEY_FOR_NIXBUILD: ${{ secrets.SSH_KEY_FOR_NIXBUILD }}
     steps:
     - uses: actions/checkout@v3
+    - run: |
+        mkdir "$HOME/.ssh"
+        chmod 700 "$HOME/.ssh"
+        cat << EOF > "$HOME/.ssh/config"
+        Host *
+          ServerAliveInterval 300
```
basvandijk, line 26:
```suggestion
          ServerAliveInterval 0
```

.github/workflows/release.yml
```diff
@@ -18,6 +18,15 @@ jobs:
       SSH_KEY_FOR_NIXBUILD: ${{ secrets.SSH_KEY_FOR_NIXBUILD }}
     steps:
     - uses: actions/checkout@v3
+    - run: |
+        mkdir "$HOME/.ssh"
+        chmod 700 "$HOME/.ssh"
+        cat << EOF > "$HOME/.ssh/config"
+        Host *
+          ServerAliveInterval 0
```
basvandijk, line 26:
```suggestion
          ServerAliveInterval 300
```

.github/workflows/release.yml
```diff
@@ -18,6 +18,15 @@ jobs:
...
+          ServerAliveCountMax 0
```
basvandijk, line 27:
```suggestion
          ServerAliveCountMax 10
```

ggreif:
This is nice, but we need a coherent plan how to promote such changes downstream. E.g. `dfinity/motoko` will need the same GHC so that we can reuse the binary cache artefacts.

nomeata:
I thought `dfinity/motoko` imports this repo as a whole, so for `ic-ref-test` it uses the same nixpkgs pin and thus the cache can be shared?

ggreif:
Hopefully. But we also have a majority of M1 Macs in the meantime which have to build locally. So that is one difference.

PR #158, opened 2023-03-29
clear certified data if canister is reinstalled

mraszyk:
Implement Interface Spec [MR](https://github.com/dfinity/interface-spec/pull/150).


basvandijk (approved):
Looks good.

PR #160, opened 2023-04-04
WIP faster WASM runtime

basvandijk:


default.nix
```diff
@@ -100,9 +147,13 @@ let
...
             (nixpkgs.haskell.lib.failOnAllWarnings
-              staticHaskellPackages.ic-hs);
+              (staticHaskellPackages.ic-hs.overrideAttrs (old: {
+                  preInstall = ''
+                    cp ${runtime_musl}/lib/libruntime.a dist/build/libruntime.a;
+                    cp ${runtime_musl}/lib/libruntime.a dist/build/libruntime-ghc9.0.2.so;
+                  '';
+                })));
       in nixpkgs.runCommandNoCC "ic-ref-dist" {
-        allowedReferences = [];
```
mraszyk, line 105: Unfortunately, there are some references left now.

cbits/core.h
michael-weigelt, line 1: Why did you have to rename all these C functions?

mraszyk: because of linking with libcrypto (or another one) that apparently also exposes all these symbols

PR #164, opened 2023-04-27
mwe/runtime

michael-weigelt:
The haskell calls into the wasm runtime carry data which has to be serialized and deserialized into rust types, and then passed to the wasm runtime. 
This MR is a step in completing this interface. 

There are also some changes that enable a vscode workflow using nix-shell, rust-analyzer and haskell language server.


src/IC/Ref/Types.hs
```diff
@@ -144,14 +144,17 @@ data Message
...
-
 type Subnet = (EntityId, SubnetType, W.Word64, SecretKey, [(W.Word64, W.Word64)])
 
 isRootSubnet :: Subnet -> Bool
 isRootSubnet (_, _, _, _, ranges) = checkCanisterIdInRanges ranges nns_canister_id
   where
     nns_canister_id = wordToId 0
 
+-- TODO: Make Subnet a record and use the accessor instead
+entityId :: Subnet -> EntityId
```
mraszyk, line 155: shouldn't this be called `subnet_id` instead to prevent clashes in the future?

src/IC/Canister.hs
```diff
@@ -123,8 +123,29 @@ timestampterm (Timestamp t) = TInteger $ fromIntegral t
...
+  ("balance", cyclesterm bal), 
+  ("status", stringterm $ show status), 
+  ("certificate", maybeterm certterm cert), 
+  ("canister_version", TInteger $ fromIntegral can_version), 
+  ("global_timer", TInteger $ fromIntegral glob_timer),
+  ("controllers", TList $ map entityterm $ toList ctrls),
+  ("memory_allocation", TInteger $ fromIntegral mem_alloc),
+  ("freeze_threshold", TInteger $ fromIntegral freeze_thresh),
+  ("subnet_id", entityterm subnet_id),
+  ("subnet_type", stringterm $ show subnet_type),
```
mraszyk, line 145: does this properly deserialize into a Rust enum?

rs/runtime/src/lib.rs
```diff
@@ -405,23 +444,24 @@ pub fn invoke(arg: &str) -> String {
...
 
+    
     let sandbox_safe_system_state = SandboxSafeSystemState::new_internal(
-        canister_id,
-        Running,                 // TODO(5): canister status
-        NumSeconds::new(1 << 5), // TODO(6): freeze_threshold
-        BestEffort,              // TODO(7): memory_allocation
-        Cycles::new(1 << 40),    // TODO(8): initial_cycles_balance
-        BTreeMap::new(),         // TODO(9): call_context_balances (only the current call context suffices)
+        env.canister_id.try_into().unwrap(),
```
mraszyk, line 449: why not `canister_id` as before?

mraszyk (approved):
LGTM

PR #167, opened 2023-05-02
parse results returned by the Rust invoke function in Haskell

mraszyk:


mraszyk:
merge into mwe/runtime2

PR #168, opened 2023-05-09
fire canister http tests in batches

mraszyk:
This PR fires canister http tests in sequential batches of 5 tests per batch so that the httpbin server is not overloaded with requests.


basvandijk (approved):

mraszyk:
Not needed after all.

PR #171, opened 2023-05-10
do not include keep-alive header in httpbin response

mraszyk:
This MR removes the keep-alive connection header from httpbin response as the httpbin server does not persist any connections.

### DIFF
--- a/src/IC/Test/Spec/HTTP.hs
+++ b/src/IC/Test/Spec/HTTP.hs
@@ -183,14 +183,13 @@ canister_http_calls sub =
 
     , simpleTestCase "small maximum possible response size" ecid $ \cid -> do
       let s = "hello_world"
-      {- Response headers (size: 136)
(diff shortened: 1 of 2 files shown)


src/IC/Test/Spec/HTTP.hs
```diff
@@ -183,82 +183,76 @@ canister_http_calls sub =
...
-          Content-Type: text/html; charset=utf-8
-          Content-Length: 12
+      {- Response headers (size: 116)
+          Content-Type: application/octet-stream
+          Content-Length: 11
           Access-Control-Allow-Origin: *
           Access-Control-Allow-Credentials: true
       -}
-      let header_size = 136
+      let header_size = 116
```
dsd-at-dfinity, line 192: ... I guess the delta is 20 and not 22 because <Key> and <Value> are measured individually in `<Key>: <Value>`.

PR #173, opened 2023-05-14
WIP: fast WASM execution engine

mraszyk:


mraszyk:
We're going to use [wasmtime-hs](https://github.com/dfinity/wasmtime-hs/) to optimize the WASM execution engine.

PR #175, opened 2023-05-20
drop nix-build-uncached

mraszyk:
The command `nix-build-uncached` does not seem to run the actual builds, so this PR replaces `nix-build-uncached` by `nix-build`.


basvandijk (approved):

nomeata:
It should build them unless the artifact is already in our nix cache. But it's a small optimization, so if it doesn't work no worries.

PR #179, opened 2023-05-22
decrease number of threads and request submission latency

mraszyk:
**Changes**

This PR makes the following changes:
- run steps of the IC on a single thread with `yield` in between consecutive steps (given that every step needs to acquire a lock on the IC state, we don't gain much by having tight loops running steps of the IC on multiple threads)
- run system tasks (heartbeats and timers) whenever an ingress message result (reply/reject) is put into the certified state (as opposed to when the user submits the request which decreases the latency of the 202 HTTP response to the user request)

**Benchmarks**

base-line (a1b3f6706c33a40f1b7faa65a9652b3ddf39d1a6)
ic-ref-test: 410.51 s
ic-nns-init: 14.3 s

this PR
ic-ref-test: 409.86 s
ic-nns-init: 11.6 s

### DIFF
--- a/src/IC/HTTP.hs
+++ b/src/IC/HTTP.hs
@@ -43,10 +43,14 @@ withApp subnets systemTaskPeriod backingFile action =
           lift getTimestamp >>= setAllTimesTo
           processSystemTasks
     loopIC :: Store IC -> IO ()
-    loopIC store = forever $ do
-        threadDelay 10000
-        modifyStore store aux
+    loopIC store = loop 1000
       where
+        loop :: Int -> IO ()
+        loop backoff = do
+          threadDelay backoff
+          b <- modifyStore store aux
+          if b then loop 1000
+          else loop (2 * backoff)
         aux = do
           lift getTimestamp >>= setAllTimesTo
           runStep
--- a/src/IC/HTTP.hs
+++ b/src/IC/HTTP.hs
@@ -3,7 +3,7 @@
 module IC.HTTP where
 
 import Network.Wai
-import Control.Concurrent (threadDelay)
+import Control.Concurrent (threadDelay, yield)
 import Control.Concurrent.Async (withAsync)
 import Network.HTTP.Types
 import qualified Data.Text as T
@@ -43,14 +43,10 @@ withApp subnets systemTaskPeriod backingFile action =
           lift getTimestamp >>= setAllTimesTo
           processSystemTasks
     loopIC :: Store IC -> IO ()
-    loopIC store = loop 1000
+    loopIC store = forever $ do
+        yield
+        modifyStore store aux
       where
-        loop :: Int -> IO ()
-        loop backoff = do
-          threadDelay backoff
-          b <- modifyStore store aux
-          if b then loop 1000
-          else loop (2 * backoff)
         aux = do
           lift getTimestamp >>= setAllTimesTo
           runStep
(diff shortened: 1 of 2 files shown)


src/IC/HTTP.hs
```diff
@@ -29,18 +29,27 @@ import IC.Utils
...
+    loopSystemTasks store = forever $ do
         threadDelay systemTaskPeriod
         modifyStore store aux
       where
         aux = do
           lift getTimestamp >>= setAllTimesTo
           processSystemTasks
+    loopIC :: Store IC -> IO ()
+    loopIC store = forever $ do
+        threadDelay 2000
```
mraszyk, line 47: Still need to understand why this constant works and if we could achieve the same effect without such a constant.

mraszyk: Unfortunately, this constant does not always work...

mraszyk: ...unless `processSystemTasks` after every incoming message is disabled.

src/IC/HTTP.hs
```diff
@@ -94,25 +103,18 @@ handle store req respond = case (requestMethod req, pathInfo req) of
   where
     runIC :: StateT IC IO a -> IO a
     runIC a = do
+      -- it is important that this thread returns, else warp is blocked somehow
       x <- modifyStore store $ do
         -- Here we make IC.Ref use “real time”
         lift getTimestamp >>= setAllTimesTo
         processSystemTasks
```
mraszyk, line 110: Commenting out this line helps with ic-nns-init, but makes some ic-ref-tests flaky.

mraszyk: The issue with ic-nns-init is due to HTTP2 being used by ic-nns-init by default.

src/IC/HTTP.hs
```diff
@@ -94,25 +103,18 @@ handle store req respond = case (requestMethod req, pathInfo req) of
   where
     runIC :: StateT IC IO a -> IO a
     runIC a = do
+      -- it is important that this thread returns, else warp is blocked somehow
```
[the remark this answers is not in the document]
mraszyk: Thank you for your assesment! The issue with ic-nns-init was due to HTTP2 being used by ic-nns-init by default.

src/IC/HTTP.hs
mraszyk: The issue with ic-nns-init is due to HTTP2 being used by ic-nns-init by default, so using a simple `yield` here.

src/IC/HTTP.hs
```diff
@@ -29,18 +29,27 @@ import IC.Utils
...
+    loopSystemTasks :: Store IC -> IO ()
+    loopSystemTasks store = forever $ do
         threadDelay systemTaskPeriod
         modifyStore store aux
       where
         aux = do
           lift getTimestamp >>= setAllTimesTo
           processSystemTasks
+    loopIC :: Store IC -> IO ()
+    loopIC store = forever $ do
```
basvandijk, line 46: It seems the previous `loopIC` implementation would terminate when `runStep` would return `False`. That's not the case anymore. Is that intended?

basvandijk (approved):

src/IC/HTTP.hs
mraszyk: Yes, that's intended because now we'd run `loopIC` on a single thread for the entire duration of running ic-ref. Before `loopIC` was spawned on a separate thread for every HTTP request received and thus could terminate if there's no more work to be done.

PR #181, opened 2023-05-24
Refactoring to support calling into Rust wasm runtime

mraszyk:


src/IC/Canister.hs
```diff
@@ -94,60 +126,107 @@ parseCanister bytes = do
...
 
-invoke :: WasmState -> CanisterEntryPoint (TrapOr r) -> TrapOr (WasmState, r)
+invoke :: CanisterSnapshot -> CanisterEntryPoint (TrapOr r) -> TrapOr (CanisterSnapshot, r)
 invoke s f =
   case perform f s of
     (_, Trap msg) -> Trap msg
     (s', Return r) -> Return (s', r)
 
+invokeRust :: RuntimeState -> EntryPoint -> IO (TrapOr (RuntimeState, UpdateResult))
+invokeRust _s _ep = error "not implemented"
```
dsd-at-dfinity, line 198: How about we give this a more descriptive name? :-)

```suggestion
initWasmBox _s _ep = error "not implemented"
```

src/IC/Canister.hs
```diff
@@ -94,60 +126,107 @@ parseCanister bytes = do
...
+invokeToCanisterActions s ep = ((\(w, r) -> (w, snd r)) <$>) <$> invokeRust s ep
+
+invokeToNoCyclesResponse :: RuntimeState -> EntryPoint -> IO (TrapOr (RuntimeState, ([MethodCall], CanisterActions)))
+invokeToNoCyclesResponse s ep = ((\(w, (a, b)) -> (w, (ca_new_calls a, b))) <$>) <$> invokeRust s ep
+
+invokeQuery :: RuntimeState -> EntryPoint -> IO (TrapOr Response)
+invokeQuery s ep = (\res -> res >>= (\(_, r) -> response $ ca_response $ fst r)) <$> invokeRust s ep
+  where
+    response Nothing = Trap "Canister did not respond."
+    response (Just r) = Return r
```
dsd-at-dfinity, line 222: If this is codified in the entry point, I wonder why we need separate functions here ... but I probably lack context.

mraszyk: It's because of the result type. `invokeWasmBox` returns all possible information and then we prune it based on the actual entry point.

PR #185, opened 2023-05-31
update Wasm validation conditions

mraszyk:
Update Wasm module validation conditions according to the Interface [specification](https://ic-interface-spec.netlify.app/docs/#system-api-module).

Note. Composite queries will be released in the specification soon so this PR already includes them in the WASM module validation conditions.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -51,6 +51,7 @@ let wabt-tests = nixpkgs.runCommandNoCC "wabt-tests" {
 } ''
   mkdir -p $out
   python3 ${wabt-definitions}/gen.py
+  cp *.wasm $out
   for f in $(ls *.wat)
   do
     wat2wasm $f -o "$out/$f.wasm"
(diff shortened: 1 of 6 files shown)


default.nix
```diff
@@ -40,12 +41,30 @@ let staticHaskellPackages = nixpkgs.pkgsStatic.haskellPackages.override {
...
+} ''
+  mkdir -p $out
+  python3 ${wabt-definitions}/gen.py
+  cp *.wasm $out
+  for f in $(ls *.wat)
+  do
+    wat2wasm $f -o "$out/$f.wasm"
+  done
+  ''; in
+
```
basvandijk, line 60: Why not just inline `${./wabt-tests}` like: 
```suggestion
let wabt-tests = nixpkgs.runCommandNoCC "wabt-tests" {
  buildInputs = [ nixpkgs.wabt python ];
  allowedRequisites = [];
} ''
  mkdir -p $out
  python3 ${./wabt-tests}/gen.py
  cp *.wasm $out
  for f in $(ls *.wat)
  do
    wat2wasm $f -o "$out/$f.wasm"
  done
  ''; in

```

basvandijk (approved):

PR #186, opened 2023-06-01
implement IC-1462

mraszyk:
This PR implements this Interface spec [PR](https://github.com/dfinity/interface-spec/pull/143) replacing some HTTP errors for error conditions in the IC protocol (not on the HTTP layer) by HTTP responses with a CBOR body describing the error.

### DIFF
--- a/src/IC/HTTP/Request.hs
+++ b/src/IC/HTTP/Request.hs
@@ -149,6 +149,11 @@ entitiyId = fmap EntityId <$> blob
 
 -- Printing responses
 response :: ReqResponse -> GenR
+response (RequestError (c, s, err)) = rec $
+    [ "error_code" =: GText (T.pack $ errorCode c) | c <- maybeToList err] ++
+    [ "reject_code" =: GNat (fromIntegral (rejectCode c))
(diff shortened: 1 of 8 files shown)


dsd-at-dfinity:
Could you please add a short PR description? :-)

src/IC/HTTP/Request.hs
```diff
@@ -149,6 +149,11 @@ entitiyId = fmap EntityId <$> blob
 
 -- Printing responses
 response :: ReqResponse -> GenR
+response (RequestError (c, s, err)) = rec $
+    [ "error_code" =: GText (T.pack $ errorCode c) | c <- maybeToList err] ++
```
dsd-at-dfinity, line 153: qq: If err is `Nothing`, the response is malformed, no?

dsd-at-dfinity (approved):

src/IC/HTTP/Request.hs
mraszyk: The error code is optional (and under-specified) so we don't need to include it.

PR #188, opened 2023-06-05
add date header to httpbin responses and make http header names lower-case

mraszyk:
This MR implements changes necessary to make the implementation of httpbin use a high-level server API (e.g., axum) that always includes a date header in its responses and produces lower-case http header names.

### DIFF
--- a/httpbin-rs/src/main.rs
+++ b/httpbin-rs/src/main.rs
@@ -46,33 +46,35 @@ impl Response {
     }
     fn set_content_length(&mut self) {
         self.headers
-            .push(("Content-Length".to_string(), self.body.len().to_string()));
+            .push(("content-length".to_string(), self.body.len().to_string()));
     }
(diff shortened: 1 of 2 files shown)


httpbin-rs/src/main.rs
```diff
@@ -46,33 +46,35 @@ impl Response {
     }
     fn set_content_length(&mut self) {
         self.headers
-            .push(("Content-Length".to_string(), self.body.len().to_string()));
+            .push(("content-length".to_string(), self.body.len().to_string()));
```
dsd-at-dfinity, line 49: Per spec, the header names are camel case. Why is this lower case?

dsd-at-dfinity (approved):

httpbin-rs/src/main.rs
mraszyk: because this is what the axum server produces (I was trying to do what the axum server implementation does)

PR #192, opened 2023-06-12
trim long blobs in json debug state

mraszyk:
The JSON debug state exposed at `http://localhost:<port>` is intended for manual inspection in a web browser and thus we should trim long blobs that slow down loading the JSON file in the web browser without facilitating human readability.

### DIFF
--- a/src/IC/Debug/JSON.hs
+++ b/src/IC/Debug/JSON.hs
@@ -37,6 +37,7 @@ import IC.Canister.Snapshot
 import IC.Canister
 import IC.Ref
 import IC.Crypto
+import IC.Hash
 
 customOptions :: Options
 customOptions = defaultOptions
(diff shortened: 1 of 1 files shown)


src/IC/Debug/JSON.hs
```diff
@@ -48,7 +49,9 @@ instance ToJSON W.Value where
     toEncoding = genericToEncoding customOptions
 
 instance ToJSON BS.ByteString where
-    toJSON = String . H.encodeHex . BS.toStrict
+    toJSON x
+      | BS.length x <= 1000 = String $ H.encodeHex $ BS.toStrict x
+      | otherwise = String $ (H.encodeHex $ BS.toStrict $ BS.take 1000 x) <> "... (sha256: 0x" <> H.encodeHex (BS.toStrict $ sha256 x) <> ")"
```
dsd-at-dfinity, line 54: Can the user really make sense of the first 1000 characters? You might have a specific use case in mind that I'm overlooking. But from mpov, I would just show the sha256 in this case. It shortens the output even more and we don't lose information. wdyt?

dsd-at-dfinity (approved):

mraszyk:
> Is anyone using this debug endpoint even? It contributes significantly to the build time, I believe (JSON instances). Maybe remove it?

We're exploring its potential for showing debug information about call contexts so I wouldn't drop it at this point.

src/IC/Debug/JSON.hs
mraszyk: Makes sense. Fixed.

PR #195, opened 2023-06-15
catch HTTP exception from HTTPS outcalls

mraszyk:
Catch HTTP exceptions from canister HTTPS outcalls and return them as reject responses with SYS_TRANSIENT reject code. Currently, ic-ref hangs upon an HTTP exception due to, e.g., the remote server being unreachable.

### DIFF
--- a/src/IC/Ref/IO.hs
+++ b/src/IC/Ref/IO.hs
@@ -19,10 +19,10 @@ import qualified Network.TLS.Extra.Cipher as C
 import qualified Network.HTTP.Client as C
 import qualified Network.HTTP.Client.TLS as C
 import Network.HTTP.Types.Status (statusCode)
+import Control.Exception
 import Data.CaseInsensitive (original)
 import Data.Default.Class (def)
 import Data.Row ((.==), (.+))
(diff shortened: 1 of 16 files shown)


src/IC/Ref/IO.hs
```diff
@@ -22,10 +22,11 @@ import Network.HTTP.Types.Status (statusCode)
 import Data.CaseInsensitive (original)
 import Data.Default.Class (def)
 import Data.Row ((.==), (.+))
+import UnliftIO.Exception
```
basvandijk, line 25: Why do you need `unliftio` here? `sendHttpRequest` is in the IO monad so the normal [`try`](https://hackage.haskell.org/package/base-4.18.0.0/docs/Control-Exception.html#v:try) should be enough no?
```suggestion
import Control.Exception
```

basvandijk (approved):

src/IC/Ref/IO.hs
mraszyk: Indeed. Thank you for the suggestion. I've dropped the `unliftio` dependency.

PR #197, opened 2023-06-16
start working on IC/Wasm/Wasmtime.hs

mraszyk:


mraszyk:
We should use `ST` instead of `IO` so the refactoring from this MR is not really useful.

PR #198, opened 2023-06-19
implement and test canister history

mraszyk:
Interface spec MR: https://github.com/dfinity/interface-spec/pull/103

### DIFF
--- a/src/IC/Ref.hs
+++ b/src/IC/Ref.hs
@@ -225,7 +225,7 @@ checkEffectiveCanisterID :: RequestValidation m => CanisterId -> CanisterId -> M
 checkEffectiveCanisterID ecid cid method arg
   | cid == managementCanisterId =
     if | method == "provisional_create_canister_with_cycles" -> pure ()
-       | method `elem` ["create_canister", "raw_rand", "http_request", "ecdsa_public_key", "sign_with_ecdsa"] ->
+       | method `elem` ["create_canister", "canister_info", "raw_rand", "http_request", "ecdsa_public_key", "sign_with_ecdsa"] ->
          throwError $ ExecutionError (RC_CANISTER_REJECT, "Management method " ++ method ++ " cannot be invoked via ingress calls", Nothing)
        | otherwise -> case Codec.Candid.decode @(R.Rec ("canister_id" R..== Principal)) arg of
                         Left err ->
@@ -456,13 +456,14 @@ processMessage m = case m of
       wasm_state <- getCanisterState callee
       can_mod <- getCanisterMod callee
       env <- canisterEnv callee
+      sender_canister_version_from_system <- getCanisterVersion callee
       invokeEntry ctxt_id callee wasm_state can_mod env entry >>= \case
         Trap msg -> do
           -- Eventually update cycle balance here
           rememberTrap ctxt_id msg
         Return (new_state, (call_actions, canister_actions)) -> do
           modCanister callee $ \cs -> cs { last_action = Just entry }
-          performCallActions ctxt_id call_actions
+          performCallActions ctxt_id call_actions (fromIntegral sender_canister_version_from_system)
           performCanisterActions callee canister_actions
           setCanisterState callee new_state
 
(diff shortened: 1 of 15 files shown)


src/IC/Ref.hs
```diff
@@ -591,28 +592,49 @@ invokeEntry ctxt_id canister_id wasm_state can_mod env entry = do
...
+  where
+    validate_subnet_message call = if (call_callee call `elem` subs) && not isOnRootSubnet then Left (RC_DESTINATION_INVALID, "Only NNS canisters can call a subnet ID directly.")
+                                   else Right ()
+    sender_canister_version_methods = ["create_canister", "update_settings", "install_code", "uninstall_code", "provisional_create_canister_with_cycles"]
+    validate_sender_canister_version call =
+      if call_callee call `elem` (managementCanisterId : subs) && call_method_name call `elem` sender_canister_version_methods then
+        case fetchSenderCanisterVersion (call_arg call) of  Nothing -> Right ()
+                                                            Just s -> if s == sender_canister_version then Right ()
+                                                                      else Left (RC_CANISTER_ERROR, "Canister must set sender_canister_version matching ic0.canister_version")
+      else Right ()
```
basvandijk, line 615: I find the above a bit hard to read, also because of the long lines. What about something more compact like:
```suggestion
validate_call call subs isOnRootSubnet sender_canister_version = do
  validate_subnet_message call
  validate_sender_canister_version call
  where
    validate_subnet_message call = 
      when (call_callee call `elem` subs) && not isOnRootSubnet) $ 
        Left (RC_DESTINATION_INVALID, "Only NNS canisters can call a subnet ID directly.")
    sender_canister_version_methods = ["create_canister", "update_settings", "install_code", "uninstall_code", "provisional_create_canister_with_cycles"]
    validate_sender_canister_version call =
      when (call_callee call `elem` (managementCanisterId : subs) && call_method_name call `elem` sender_canister_version_methods) $
        case fetchSenderCanisterVersion (call_arg call) of 
          Nothing -> Right ()
          Just s ->
            unless (s == sender_canister_version) $
              Left (RC_CANISTER_ERROR, "Canister must set sender_canister_version matching ic0.canister_version")
```

src/IC/Ref/Management.hs
```diff
@@ -88,13 +90,32 @@ invokeManagementCanister _ _ _ Closure{} = error "closure invoked on management
 invokeManagementCanister _ _ _ Heartbeat = error "heartbeat invoked on management canister"
 invokeManagementCanister _ _ _ GlobalTimer = error "global timer invoked on management canister"
 
+icAddCanisterHistory :: ICM m => CanisterId -> CallId -> Maybe W.Word64 -> ChangeDetails -> m ()
+icAddCanisterHistory canister_id ctxt_id sender_canister_version change_details = do
+    can_state <- getCanister canister_id
+    let timestamp_nanos = case time can_state of Timestamp t -> t
```
basvandijk, line 96: Why a case and not:
```suggestion
    let Timestamp timestamp_nanos = time can_state
```

src/IC/Ref/Management.hs
```diff
@@ -272,19 +303,25 @@ icUninstallCode r = do
...
+icUpdateCanisterSettings ctxt_id r = do
     let canister_id = principalToEntityId (r .! #canister_id)
     validateSettings (r .! #settings)
     applySettings canister_id (r .! #settings)
     bumpCanisterVersion canister_id
+    case ((r .! #settings) .! #controllers) of
+      Nothing -> return ()
+      Just _ -> do
+        new_controllers <- S.toList <$> getControllers canister_id
+        icAddCanisterHistory canister_id ctxt_id (r .! #sender_canister_version) (ControllersChange new_controllers)
```
basvandijk, line 324: You could consider using [`for_`](https://hackage.haskell.org/package/base-4.18.0.0/docs/Data-Foldable.html#v:for_) here like:
```suggestion
    for_ ((r .! #settings) .! #controllers) $ \_ -> do
        new_controllers <- S.toList <$> getControllers canister_id
        icAddCanisterHistory canister_id ctxt_id (r .! #sender_canister_version) (ControllersChange new_controllers)
```
but that's also a matter of taste so I'll leave it up to you ;).

basvandijk (approved):
Looks great. I added some styling remarks.

src/IC/Ref/Management.hs
mraszyk: Thank you for pointing out this pattern. Applied using `forM`.

src/IC/Ref.hs
mraszyk: Makes sense! Simplified the code according to your suggestion.

PR #199, opened 2023-06-19
reproduce wasmtime-hs segfault

mraszyk:
Steps to reproduce:

1. Store this minimal WASM to a file:
```
$ echo -e "00000000  00 61 73 6d 01 00 00 00                           |.asm....|\n00000008" | xxd -r > e.wasm
$ hexdump -C e.wasm
00000000  00 61 73 6d 01 00 00 00                           |.asm....|
00000008
```
create an empty `counter.did` file
```
$ touch counter.did
```
and create a `dfx.json` file with the following content:
```
$ cat dfx.json
{
  "canisters": {
    "counter": {
      "type": "custom",
      "build": "",
      "wasm": "",
      "candid": "counter.did"
    }
  }
}
```

2. Run `ic-ref`:
```
$ cabal run ic-ref -- --listen-port 8080
Starting ic-ref...
Running at http://127.0.0.1:8080/
```

3. Create and install a canister with the WASM from step 1:
```
$ rm -rf .dfx && dfx canister create counter --no-wallet && dfx canister install counter --wasm e.wasm
```
then we see the following output from `ic-ref`:
```
done
Segmentation fault (core dumped)
```

### DIFF
--- a/src/IC/Canister.hs
+++ b/src/IC/Canister.hs
@@ -21,7 +21,6 @@ import qualified Data.ByteString.Lazy.Char8 as BS
 import Data.Char (chr)
 import Data.List
 import Control.Monad
-import Control.Monad.ST
 import Data.Foldable
 import Control.Monad.Except
 import Codec.Compression.GZip (decompress)
@@ -38,7 +37,6 @@ import IC.Canister.Snapshot
 import IC.Canister.Imp
 import IC.Hash
 import IC.Utils
-import UnliftIO.Exception
 
 -- Here we can swap out the purification machinery
 type WasmState = CanisterSnapshot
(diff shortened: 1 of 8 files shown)


src/IC/Canister.hs
```diff
@@ -98,8 +104,24 @@ parseCanister bytes = do
...
+          engine <- newEngine
+          store <- newStore engine
+          ctx <- storeContext store
+          myModule <- case newModule engine (unsafeFromByteString $ BS.toStrict decodedModule) of
+                            Left err -> putStrLn (show err) >> error ""
+                            Right r -> return r
+          -- the following crashes
+          f <- newFunc ctx msg_reply
+          r <- newInstance ctx myModule (Vec.fromList [toExtern f])
+          -- the following "hangs" (never prints "done")
```
mraszyk, line 117: probably expected though because `newInstance` is expected to fail if we don't provide any host function for the `msg_reply` import and if we don't catch the exception (compare with the previous commit on this branch catching all exceptions where we'd actually see an error message and then get a proper reject message later from Winter)

mraszyk:
Note: I don't observe a crash when running with `valgrind`. I still observe a crash when running with `gdb` though.

mraszyk:
The issue has been fixed in wasmtime-hs: https://github.com/dfinity/wasmtime-hs/pull/31

PR #211, opened 2023-07-14
synchronize node state heights before making non-replicated calls in ic-ref-test

mraszyk:
When running the compliance tests against a BN and a testnet with multiple nodes, we need to make sure that the node states are up-to-date before making non-replicated calls (query calls and read state requests). Otherwise, we might hit a node with an old state and spuriously fail when making a query call against that node.

Hence, this PR synchronizes the node states before query calls and read state requests as follows:
- it determines the subnet to which the requested will be routed
- it queries the certified times of all nodes on that subnet (using read state requests)
- if the times are not all equal, it queries the certified times of all nodes on that subnet in a loop until the maximum time from the initial certified times (obtained in the previous step) is at most the minimum time from the current certified times (obtained in the current step) => once this is the case, then all nodes have certainly caught up


src/IC/Test/Agent.hs
```diff
@@ -416,10 +417,41 @@ postCBOR path gr = do
...
+sync_height cid = mapM (\(ns, rs) -> do
+  let ranges = map (\(a, b) -> (wordToId' a, wordToId' b)) rs
+  when (any (\(a, b) -> a <= cid && cid <= b) ranges) $ do
+    hs <- get_heights ns
+    let h = maximum hs
+    unless (length (nub hs) <= 1) $
+      waitFor $ do
+        hs <- get_heights ns
+        return $ h <= minimum hs
+    return ()) subnets
```
basvandijk, line 450: I would use a `forM` instead of a `mapM` here so that you don't need to wrap that big function in parenthesis. You also shouldn't need the final `return ()` I think:
```suggestion
sync_height cid = forM subnets $ \(ns, rs) -> do
  let ranges = map (\(a, b) -> (wordToId' a, wordToId' b)) rs
  when (any (\(a, b) -> a <= cid && cid <= b) ranges) $ do
    hs <- get_heights ns
    let h = maximum hs
    unless (length (nub hs) <= 1) $
      waitFor $ do
        hs <- get_heights ns
        return $ h <= minimum hs
```

basvandijk (approved):

src/IC/Test/Agent.hs
mraszyk: Thank you for the suggestion! Applied.

src/IC/Test/Agent.hs
```diff
@@ -209,12 +212,13 @@ data AgentConfig = AgentConfig
     { tc_root_key :: Blob
     , tc_manager :: Manager
     , tc_endPoint :: String
+    , tc_subnets :: [([String], [(W.Word64, W.Word64)])]
```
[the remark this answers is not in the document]
mraszyk: Added a new type
```
data AgentSubnetConfig = AgentSubnetConfig
  { tc_node_addresses :: [String]
  , tc_canister_ranges :: [(W.Word64, W.Word64)]
  }
```
to the `IC.Agent` module.

basvandijk:
Looks great now!

